PR #7, opened 2015-08-25
Update README.md to reflect release 1.2.0

sbennett1990:
Update the URLs for the `scrypt` source tarball and the GPG signature, since these changed with the 1.2.0 release.


cperciva:
Thanks!

PR #8, opened 2015-08-27
some autotools yak shaving

barak:


cperciva:
Thanks for all the cleanups -- as you may have guessed, I'm really not an autotools guy.  I'm happy with most of this but I've left comments on a few patches.  Can you update this to reflect those comments?

barak:
Sure, will tweaks the tweaks.

Regarding version numbers, there is actually a git command for getting a reasonable version number to use in an executable generated from a repo snapshot:
````
$ git describe --always --tags --dirty
1.2.0-17-g46feac9
````

barak:
For some background: these tweaks are just forward-ports of changes I made to get scrypt to compile more happily on the Debian autobuilders, which like to pass odd flags, use strange compilers, rebuild everything from source, etc.

They pass --disable-maintainer-mode anyway, so that wasn't actually an issue. I removed it on autopilot. Just removed that patch. Will take care of the other issues as well.

Regarding the one true version number, how about having ./configure generate the scrypt-version file using information in configure.ac, instead of having autoconf generate the configure file using information in scrypt-version?

barak:
ok, i pushed the version patches to a different branch

cperciva:
Good to know the background, thanks!

With regard to version numbers, I'm not yet 100% convinced that I want to force people to have git installed; and it's much easier for my release-tarball-building if I have the version number in a separate file rather than updating configure.ac.

cperciva:
Thanks!

PR #9, opened 2015-09-02
fix typo in comment

ben0x539:
The warning comment added in ccb623b6d refers to `crypt_scrypt` instead of `crypto_scrypt`.


cperciva:
Thanks!  This is worth $1 of tarsnap bug bounty if you want it.

PR #10, opened 2015-09-27
fclose() infile and outfile

gperciva:
I hope I'm not missing anything silly here.  I think we want to `fclose()` the files we opened, even if those files are `stdin` and `stdout`.

I considered putting this above the `/* If we failed, print the right error message and exit. */` since closing `stdout` wouldn't harm the `warnp()` and `warn0()` because they print to `stderr`, but then I decided I was was being insufficiently paranoid so it would be better to move the clean-up to the very bottom of `main()`.

### DIFF
--- a/main.c
+++ b/main.c
@@ -135,6 +135,12 @@ main(int argc, char *argv[])
 	insecure_memzero(passwd, strlen(passwd));
 	free(passwd);
 
+	/* Close any files we opened. */
+	if (infile != stdin)
+		fclose(infile);
+	if (outfile != stdout)
+		fclose(outfile);
+
 	/* If we failed, print the right error message and exit. */
 	if (rc != 0) {
 		switch (rc) {


cperciva:
There's a few ways to look at this:

1. We should close any files we used.

2. We don't really need to close these files, because stdio closes all open file handles when it exits.

3. We should close any files we opened, in case this turns into a function which is called many times in the future and we lose the file-handle-autoclosing.

So I'd go with closing the files, but only if they're not stdin/stdout.

gperciva:
Ok.  Would you object if I added an

    int infile_not_stdin;
    int outfile_not_stdout;

?  I think that would make the code more readable.  The current PR doesn't do that; it repeats the `if (strcmp...` and `(argc > 1)` lines.

cperciva:
Just compare the file handles.  "if (infile != stdin)" etc.

gperciva:
wow, that makes an uncommon amount of sense for C code!   :)    PR updated.

main.c
```diff
@@ -135,6 +135,14 @@ main(int argc, char *argv[])
 	insecure_memzero(passwd, strlen(passwd));
 	free(passwd);
 
+	/* Close files if they are not streams. */
```
cperciva: All files are streams.  What you mean is "Close any files we opened".

I'd skip the braces for such trivial code btw, but I don't mind if you prefer to keep them in.

gperciva:
updated

Issue #12, opened 2015-09-27
Cannot compile tests with clang 3.4.1: uint8_t vs. char


gperciva:
So `crypto_scrypt()` is careful to always use `uint8_t`, whereas `test/main.c` uses `char`.

I can sprinkle a bunch of `(uint_t*)` in `test/main.c`, but then `strlen()` isn't happy with the result.  Technically, `char` isn't guaranteed to be `uint8_t`, so I'm not certain about the most portable way to fix this.  I suppose we could add the string lengths (hard-coded) into `static struct scrypt_test`, but that feels like a hack.

```
td@gin: ~/src/scrypt/test (compile-tests)
$ clang --version
Ubuntu clang version 3.4-1ubuntu3 (tags/RELEASE_34/final) (based on LLVM 3.4)
Target: x86_64-pc-linux-gnu
Thread model: posix
td@gin: ~/src/scrypt/test (compile-tests)
$ CC=clang bmake
clang -O2 -Wno-error -I .. -DCONFIG_H_FILE=\"config_freebsd.h\" -I ../libcperciva/alg -I ../libcperciva/cpusupport -I ../lib/crypto -I ../libcperciva/util -I . -D CPUSUPPORT_CONFIG_FILE=\"cpusupport-config.h\"  -Wall -Wstrict-prototypes -Wmissing-prototypes -Wpointer-arith -Wno-uninitialized -Wreturn-type -Wcast-qual -Wpointer-arith -Wwrite-strings -Wswitch -Wshadow -Werror   -c main.c
main.c:35:17: error: passing 'const char *' to parameter of type
      'const uint8_t *' (aka 'const unsigned char *') converts between pointers
      to integer types with different sign [-Werror,-Wpointer-sign]
                crypto_scrypt(test->passwd, strlen(test->passwd),
                              ^~~~~~~~~~~~
../lib/crypto/crypto_scrypt.h:44:34: note: passing argument to parameter here
int crypto_scrypt(const uint8_t *, size_t, const uint8_t *, size_t, uint64_t,
                                 ^
main.c:36:7: error: passing 'const char *' to parameter of type
      'const uint8_t *' (aka 'const unsigned char *') converts between pointers
      to integer types with different sign [-Werror,-Wpointer-sign]
                    test->salt, strlen(test->salt),
                    ^~~~~~~~~~
../lib/crypto/crypto_scrypt.h:44:59: note: passing argument to parameter here
[...]

PR #13, opened 2015-09-28
Compile tests

gperciva:
fixed with clang on ubuntu 14.04, but not gcc yet.


gperciva:
addresses #12.

gperciva:
updated with const.

gperciva:
adjusted code and commit subject

PR #17, opened 2015-10-14
Replace version numbers in MAN pages

gperciva:


cperciva:
Don't rename the file -- people should be able to check out the tree and build/install.

gperciva:
huh, cute little WTF: debian's man version 2.6.7.1 will automatically replace `@DATE@` with the current date.  So if the installed `scrypt.1` contains `.Dd @DATE` then it'll update every day.

I'm at a loss as to why this would ever be desirable, but googling for "man date" or "man date replacement" doesn't provide any hints and merely seems to make google question my sexual orientation.  (why, oh why, did google stop allowing exact searches?  I mean, it could be hidden on a google.com/yes-I-am-a-techie-and-this-punctuation-is-important.html page, but instead they removed it entirely.  Growl!)

gperciva:
Update: apparently any non-date is converted to a date.  It'll replace "foo" or "this is my string" with the current date.  Text like "1 month ago" or "1 day ago" is retained unchanged.

More fun stuff: "Easter Monday 2012" is kept, as is "Eastery Monday 2012" and "Westerly Monday 2012".  "Test 1234" is replaced, but "Test test 1234" is kept.  "Replace date string" is kept, whereas "Replace date string please" is not.  So I guess it's looking for a string containing 3 words?


Anyway, I've updated this PR to keep the original `scrypt.1` filename.  At the moment, somebody installing directly from git will get an auto-updating man page date.  We could make that "Replace date string" or "Fake date string" to avoid this behaviour, but I'm not certain that's much better.

cperciva:
It's not so much that non-dates are converted to dates; rather, it's that `.Dd <something which is not accepted as a date>` is ignored, and if there isn't a valid .Dd command then today's date gets used instead.  Different parsers have different ideas of what gets accepted as a date; yours is clearly on the less strict side.

This tripped me up about a dozen years ago when I was first writing FreeBSD Update, incidentally -- there was a catpage which was being compiled differently every day, and it turned out to be because someone spelled a month name wrong and this resulted in the date not being accepted.

gperciva:
TIL "catpage" is a thing.  I mean, other than the resting state of the internet.  (try a google search for "catpage")

cperciva:
Back in the days when rendering man pages took at lot of CPU time, it made sense to cache the rendered pages -- named "catpages" because you could read them using "cat" instead of "man".  There was even a cron job which looked at the access times for the rendered files and would remove ones which hadn't been read recently in order to save disk space.

Then disk space got cheap and OS vendors decided that it made sense to ship catpages with the release.  Some time around this point the manpage and catpage files started being gzipped, too.

Then CPU cycles got cheap, and now people generally don't bother shipping catpages.

Issue #5, opened 2015-08-05
Feature request: Add verbose (-v) option 


coderb:
It would be nice if there was a -v option that showed the parameters and/or memory/cpu settings chosen when scrypting a file.

cperciva:
You're not the only person asking for this either.  Right now there is code for printing those settings but it's hidden behind #ifdef DEBUG.

I'm not going to have time to implement this in the near future, but it should be pretty straightforward and I'm always happy to review pull requests.

gperciva:
Would you object to `-v` being "version" and `-V` being "verbose"?

cperciva:
Ick, yes.  If there's an option which means "verbose" it should always be -v.  Is there a standard command line option for printing a version number aside from --version?  (We could do that, but it seems a waste to pull in getopt_long just for that...)

gperciva:
I don't know of any standard short form of `--version`.

One advantage of using `getopt_long` is that then we could offer `--maxmem`, `maxmemfrac` and the like as well, which is easier to understand than `-M` and `-m`.

gperciva:
Incidently, `spipe` and `spiped` both use 'v' for "print version number".  I guess you only accepted that PR because you didn't want `getopt_long` in there?

cperciva:
Partly because I didn't want getopt_long, partly because we didn't have a "verbose" option for spiped (nor am I really sure what it would do).

Speaking of long options though, I might write an insane getopt_long replacement (see https://github.com/Tarsnap/libcperciva/issues/9).  Let me decide if I want to do that before we decide about option names here.

cperciva:
This is now done.

PR #18, opened 2015-10-22
Command line verbose

gperciva:
First draft; addresses #5.

### DIFF
--- a/lib/scryptenc/scryptenc.c
+++ b/lib/scryptenc/scryptenc.c
@@ -30,6 +30,7 @@
 
 #include <errno.h>
 #include <fcntl.h>
+#include <inttypes.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
@@ -51,12 +52,28 @@
 #define ENCBLOCK 65536
 
 static int pickparams(size_t, double, double,
-    int *, uint32_t *, uint32_t *);
-static int checkparams(size_t, double, double, int, uint32_t, uint32_t);
+    int *, uint32_t *, uint32_t *, int);
+static int checkparams(size_t, double, double, int, uint32_t, uint32_t, int);
+
+static void
+display_params(int logN, uint32_t r, uint32_t p, size_t memlimit,
+    double opps, double maxtime)
+{
+	uint64_t N = (uint64_t)(1) << logN;
+	uint64_t mem_minimum = 128 * r * N;
+	double expected_seconds = 4 * N * p / opps;
+
+	fprintf(stderr, "Parameters used: N = %" PRIu64 "; r = %" PRIu32
+	    "; p = %" PRIu32 ";\n", N, r, p);
+	fprintf(stderr, "    This requires at least %" PRIu64 " bytes "
+	    "of memory (%zu available),\n", mem_minimum, memlimit);
+	fprintf(stderr, "    and will take approximately %.1f seconds "
+	    "(limit: %.1f seconds).\n", expected_seconds, maxtime);
+}
 
 static int
 pickparams(size_t maxmem, double maxmemfrac, double maxtime,
-    int * logN, uint32_t * r, uint32_t * p)
+    int * logN, uint32_t * r, uint32_t * p, int verbose)
 {
 	size_t memlimit;
 	double opps;
@@ -112,10 +129,8 @@ pickparams(size_t maxmem, double maxmemfrac, double maxtime,
 		*p = (uint32_t)(maxrp) / *r;
 	}
 
-#ifdef DEBUG
-	fprintf(stderr, "N = %zu r = %d p = %d\n",
-	    (size_t)(1) << *logN, (int)(*r), (int)(*p));
-#endif
+	if (verbose)
+		display_params(*logN, *r, *p, memlimit, opps, maxtime);
 
 	/* Success! */
 	return (0);
@@ -123,7 +138,7 @@ pickparams(size_t maxmem, double maxmemfrac, double maxtime,
 
 static int
 checkparams(size_t maxmem, double maxmemfrac, double maxtime,
-    int logN, uint32_t r, uint32_t p)
+    int logN, uint32_t r, uint32_t p, int verbose)
 {
 	size_t memlimit;
 	double opps;
@@ -153,6 +168,9 @@ checkparams(size_t maxmem, double maxmemfrac, double maxtime,
 	if ((opslimit / N) / (r * p) < 4)
 		return (10);
 
+	if (verbose)
+		display_params(logN, r, p, memlimit, opps, maxtime);
+
 	/* Success! */
 	return (0);
 }
@@ -160,7 +178,7 @@ checkparams(size_t maxmem, double maxmemfrac, double maxtime,
 static int
 scryptenc_setup(uint8_t header[96], uint8_t dk[64],
     const uint8_t * passwd, size_t passwdlen,
-    size_t maxmem, double maxmemfrac, double maxtime)
+    size_t maxmem, double maxmemfrac, double maxtime, int verbose)
 {
 	uint8_t salt[32];
 	uint8_t hbuf[32];
@@ -175,7 +193,7 @@ scryptenc_setup(uint8_t header[96], uint8_t dk[64],
 
 	/* Pick values for N, r, p. */
 	if ((rc = pickparams(maxmem, maxmemfrac, maxtime,
-	    &logN, &r, &p)) != 0)
+	    &logN, &r, &p, verbose)) != 0)
 		return (rc);
 	N = (uint64_t)(1) << logN;
 
@@ -214,7 +232,7 @@ scryptenc_setup(uint8_t header[96], uint8_t dk[64],
 static int
 scryptdec_setup(const uint8_t header[96], uint8_t dk[64],
     const uint8_t * passwd, size_t passwdlen,
-    size_t maxmem, double maxmemfrac, double maxtime)
+    size_t maxmem, double maxmemfrac, double maxtime, int verbose)
 {
 	uint8_t salt[32];
 	uint8_t hbuf[32];
@@ -245,7 +263,8 @@ scryptdec_setup(const uint8_t header[96], uint8_t dk[64],
 	 * key derivation function can be computed within the allowed memory
 	 * and CPU time.
 	 */
-	if ((rc = checkparams(maxmem, maxmemfrac, maxtime, logN, r, p)) != 0)
+	if ((rc = checkparams(maxmem, maxmemfrac, maxtime, logN, r, p,
+	    verbose)) != 0)
 		return (rc);
 
 	/* Compute the derived keys. */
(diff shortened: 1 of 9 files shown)


gperciva:
I know you said that you wanted to look at https://github.com/Tarsnap/libcperciva/issues/9 (weird black magick for getopt), but I figured that I might as well get this part done now.

cperciva:
This looks like a first start, but can you add some more code to pickparams and checkparams to indicate the limits we detected?  "Selecting parameters to use up to X salsa20/8 cores and up to Y bytes of memory", say.

gperciva:
Hmm, I'm not certain if this is doing the right thing.  The memory is plausible, but (as a non-cryptographer) the number of salsa20/8 cores looks really high.  Skimming the wikipedia page suggests that each core only takes 4-14 cycles, though, so I guess it *should* be a very high number.

Is it expected that the estimated memory and cores differ between encryption and decryption?  Memory differs by a power of 4, but the number of cores differs by a factor of almost 62.

```
$ ./scrypt enc -v config.log foo.txt
Please enter passphrase: 
Please confirm passphrase: 
Parameters used: N = 262144 r = 8 p = 1
    This requires up to 1035043328 (1.04e+09) bytes of memory,
    and up to 11699541 (1.17e+07) salsa20/8 cores.
td@gin: ~/src/scrypt/build (command-line-verbose)
$ ./scrypt dec -v foo.txt bar.txt
Please enter passphrase: 
Parameters used: N = 262144 r = 8 p = 1
    This requires up to 4140173312 (4.14e+09) bytes of memory,
    and up to 720051753 (7.20e+08) salsa20/8 cores.
```

gperciva:
I was also tempted to state the memory in GB; seeing `0.25 GB` means a lot more to most people than `258760832 bytes`.  Would you accepted that (in parentheses)?

cperciva:
When running `scrypt dec` the default is to allow up to 300s and up to 1/2 of "available" memory.  When running `scrypt enc` those are reduced to 5s and 1/8 of available memory, to ensure that inaccuracy in the performance measurement won't result in files which can't be decrypted on the system where they were encrypted.

For `scrypt dec` we should display the actual amount of memory and cpu usage, not the maximum allowed.

And yes, feel free to put a humansized amount into parentheses.

gperciva:
Updated PR with humansize.  Sample output:

```
$ ./scrypt enc -v config.log foo.txt
Please enter passphrase: 
Please confirm passphrase: 
Parameters used: N = 262144 r = 8 p = 1
    This requires up to 1035037184 (1.0 GB) bytes of memory,
    and up to 12074275 (12 M) salsa20/8 cores.
td@gin: ~/src/scrypt/build (command-line-verbose)
$ ./scrypt dec -v foo.txt bar.txt
Please enter passphrase: 
Parameters used: N = 262144 r = 8 p = 1
    This requires up to 4140148736 (4.1 GB) bytes of memory,
    and up to 816947403 (816 M) salsa20/8 cores.
```

QUESTION 1: I wanted to make sure that the `opslimit` would be displayed nicely, so even though I'm using humansize(), I strip the final "B" so that it displays a number without units (in bytes).  I made that code handle the case of "999 B" => "999", but of course `opslimit` has a minimum of 32768.  Should I leave this case in the code, or remove it because it will never happen in this particular example?

QUESTION 2: I don't understand what you mean by "For `scrypt dec` we should display the actual amount of memory and cpu usage, not the maximum allowed.".  I assume that you do not want me adding benchmarking and profiling code to `crypto_scrypt`.  At the moment, it displays the `memlimit` that results from `memtouse(maxmem, maxmemfrac, &memlimit)` called from `checkparams`.  If that is not correct, do you want me to display the value of `128*r*N` (which, from skimming `crypto_scrypt` appears to be the largest `posix_memalign` call)?  or rather, display the value of `128*r*N + 128*r*p + 256*r + 64`?  if so, I'm not certain how to calculate the "actual amount of CPU usage" without timing it (or using more esoteric profiling code), and surely the main point is to display this info *before* `scrypt dec` finishes (or begins) running, so that the user has a hint as to why it fails (if it's failing).

cperciva:
For `scrypt dec` we should definitely report the value `128*r*N`, since this is the amount of memory which would be required -- that way the user can say "wait, it says I need 1 GB of RAM and I don't have enough available, but I know I have way more than that".  My inclination is to report the corresponding "using X salsa8/20 cores" as well, even though it won't be very useful.

Actually, maybe report both requirement and availability... something like "This will use 1 GB of memory (3.4 GB available) and take 0.83 seconds (limit: 60 seconds)".

gperciva:
Assuming that you're suggesting that I estimate the time via `opslimit / opps`, it appears that `scryptenc_cpuperf` isn't very accurate:

```
$ time ./scrypt enc -v config.log foo.txt
Please enter passphrase: 
Please confirm passphrase: 
opps: 2.41178e+06
opslimit / opps: 5
Parameters used: N = 262144 r = 8 p = 1
    This requires at least 268 MB bytes of memory (1.0 GB available),
    and will take 5.0 seconds (limit: 5.0 seconds).

real	0m1.157s
user	0m0.780s
sys	0m0.028s
```

That's including time for me to enter a passphrase (just "a" in this case, so still fast).  How about I give the number of salsa8/20 cores, and the max time?  (leaving out any "estimated time")

cperciva:
"estimate the time via opslimit / opps"

Err, opslimit is computed as <maximum time> * opps.  As the name suggests, that's the *limit*.

To estimate the time which will actually be taken, compute `4*N*r*p / opps`.

gperciva:
Thanks, that makes more sense.

gperciva:
Based on empirical evidence, are you certain you don't mean `N*r*p / opps` ?  (without the factor of `4`?)   I don't see any multithreaded CPU activity, but it's almost exactly correct if I omit the 4.

cperciva:
What does the code say?

gperciva:
Code does not agree with wall-clock.  I've investigated the timing a bit on my intel core i5-2415M, 64-bit Ubuntu 14.04.

original `lib/scryptenc/scryptenc_cpuperf.c`:
```
512 salsa20/8 cores performed in 0.000223 seconds
opps is: 2.29313e+06
```

Artificially changing the timing, a few values:
```
artifically setting timing resolution resd: 0.1
358400 salsa20/8 cores performed in 0.100008 seconds
opps is: 3.58372e+06

artifically setting timing resolution resd: 1.0
7973376 salsa20/8 cores performed in 1.000039 seconds
opps is: 7.97307e+06

artifically setting timing resolution resd: 10
87677440 salsa20/8 cores performed in 10.000009 seconds
opps is: 8.76774e+06

artifically setting timing resolution resd: 60
519414272 salsa20/8 cores performed in 60.000057 seconds
opps is: 8.6569e+06
```

This explains the timing discrepancy I noted ("roughly a factor of 4"; this experiment shows 2.3 to 8.7).  Particularly since when I wasn't certain about timing issues, I increased the time-limit in order to have the time being dominated by the raw math & CPU, rather than program loading / buffering / etc.

If you say that I should ignore this and just print the estimated time based on the original short-term opps, I'll go ahead and do that.  Or if we want to investigate timing estimation further (since computers now are much faster than in 2009?  -- although my desktop is from 2011, so this isn't a great experiment about modern CPU speeds), I'm of course happy to do that.

gperciva:
In offline communication, Colin said that using the estimate from `opps` is fine.

New output:
```
td@gin: ~/src/scrypt/build (command-line-verbose)
$ ./scrypt enc -v config.log foo.txt
Please enter passphrase: 
Please confirm passphrase: 
Parameters used: N = 262144 r = 8 p = 1
    This requires at least 268 MB bytes of memory (1.0 GB available),
    and will take approximately 2.1 seconds (limit: 5.0 sceonds).
td@gin: ~/src/scrypt/build (command-line-verbose)
$ ./scrypt dec -v foo.txt bar.txt
Please enter passphrase: 
Parameters used: N = 262144 r = 8 p = 1
    This requires at least 268 MB bytes of memory (4.1 GB available),
    and will take approximately 1.5 seconds (limit: 300.0 sceonds).
td@gin: ~/src/scrypt/build (command-line-verbose)
```

lib/scryptenc/scryptenc.c
```diff
@@ -51,12 +53,32 @@
...
+	long int mem_minimum = 128*r*N;
+	char * human_memlimit = humansize(memlimit);
+	char * human_mem_minimum = humansize(mem_minimum);
+
+	fprintf(stderr, "Parameters used: N = %zu r = %d p = %d\n",
+	    N, (int)r, (int)p);
+	fprintf(stderr, "    This requires at least %s bytes of memory "
+	    "(%s available),\n", human_mem_minimum, human_memlimit);
+	fprintf(stderr, "    and will take approximately %.1f seconds "
+	    "(limit: %.1f sceonds).\n", 4*N*4*p / opps, maxtime);
```
cperciva: `sceonds` ?

`4*n*4*p` ?  Also, whitespace.

lib/scryptenc/scryptenc.c
```diff
@@ -51,12 +53,32 @@
...
-static int checkparams(size_t, double, double, int, uint32_t, uint32_t);
+    int *, uint32_t *, uint32_t *, int);
+static int checkparams(size_t, double, double, int, uint32_t, uint32_t, int);
+
+static void
+display_params(int logN, uint32_t r, uint32_t p, size_t memlimit,
+    double opps, double maxtime)
+{
+	size_t N = (size_t)(1) << logN;
+	long int mem_minimum = 128*r*N;
```
cperciva: Whitespace.

lib/scryptenc/scryptenc.c
```diff
@@ -112,10 +134,9 @@ pickparams(size_t maxmem, double maxmemfrac, double maxtime,
 		*p = (uint32_t)(maxrp) / *r;
 	}
 
-#ifdef DEBUG
-	fprintf(stderr, "N = %zu r = %d p = %d\n",
-	    (size_t)(1) << *logN, (int)(*r), (int)(*p));
-#endif
+	if (verbose)
+		display_params(*logN, (int)(*r), (int)(*p), memlimit,
```
cperciva: Why the casts to (int) ?

gperciva:
Updated.  Here's the current output:

```
td@gin: ~/src/scrypt/build (command-line-verbose)
$ time ./scrypt enc -v config.log foo.txt
Please enter passphrase: 
Please confirm passphrase: 
Parameters used: N = 262144 r = 8 p = 1
    This requires at least 268 MB bytes of memory (1.0 GB available),
    and will take approximately 0.5 seconds (limit: 5.0 seconds).

real	0m1.600s
user	0m0.770s
sys	0m0.040s
td@gin: ~/src/scrypt/build (command-line-verbose)
$ ./scrypt dec -v foo.txt bar.txt
Please enter passphrase: 
Parameters used: N = 262144 r = 8 p = 1
    This requires at least 268 MB bytes of memory (4.1 GB available),
    and will take approximately 0.5 seconds (limit: 300.0 seconds).
td@gin: ~/src/scrypt/build (command-line-verbose)
```

gperciva:
Fixed with `PRIu32` and `PRIu64` everywhere (I hope).

gperciva:
Also, slightly different output: I added semicolons.

```
td@gin: ~/src/scrypt/build (command-line-verbose)
$ ./scrypt enc -v config.log foo.txt
Please enter passphrase: 
Please confirm passphrase: 
Parameters used: N = 262144; r = 8; p = 1;
    This requires at least 268 MB bytes of memory (1.0 GB available),
    and will take approximately 0.4 seconds (limit: 5.0 seconds).
td@gin: ~/src/scrypt/build (command-line-verbose)
$ ./scrypt dec -v foo.txt bar.txt
Please enter passphrase: 
Parameters used: N = 262144; r = 8; p = 1;
    This requires at least 268 MB bytes of memory (4.1 GB available),
    and will take approximately 0.4 seconds (limit: 300.0 seconds).
td@gin: ~/src/scrypt/build (command-line-verbose)
```

gperciva:
Slightly refactored update (I decided it didn't make sense to separate the command-line handling from the printing code).

PR #22, opened 2015-12-03
Fixed test functionality during make

bigendiansmalls:
test_scrypt will now compile in the test directory during the initial make.

Modified file names in /test and added a Makefile.am in /test
Also changed Makefile.am in the main directory to include it
and edited README to update this.

Tested on linux and Mac OSX.


bigendiansmalls:
That should address the items you mentioned, couple were just mistakes or typos on my part.  Let me know if you'd like me to tweak anything else.  Cheers.

gperciva:
Hi bigendiansmalls,

This modification doesn't work with an out-of-tree build.  For example, from the main scrypt git repo, try:

    mkdir build/
    cd build/
    ../configure
    make

and you'll soon get:

    ../../test/test_scrypt.c:4:27: fatal error: crypto_scrypt.h: No such file or directory
     #include "crypto_scrypt.h"

I don't have time to investigate much further right now, but if you'd like, I could help fix this later in the afternoon or sometime tomorrow.

(hmm, at first glance, I'm really suspicious about removing `$(top_srcdir)` from `test/Makefile.am`)

cperciva:
I was going to ask about that too.  I think `autoreconf -i` does what all the separate commands do.

gperciva:
Right, looks like `autoreconf -i` is what to use: "autoreconf runs autoconf, autoheader, aclocal, automake, libtoolize, and autopoint (when appropriate) repeatedly to update the GNU Build System"  https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/autoreconf-Invocation.html

The `-i` installs missing aux files.

bigendiansmalls:
Re - out of tree - right - I tried it with top_srcdir - but got some very strange results where it built that directory inside test (e.g.  test/$(srcdir)/lib  was built)   that didn't seem like normal behavior (despite that it worked) -- if you try the commit before I removed it; see if it works for you.  I will look into the correct make var for that; seemed like top_srcdir was correct, but something wasn't right with it.  Let me see what I can find.

bigendiansmalls:
Let me try the autoreconf -i on my 2 test boxes - I'm by no means a GNU packing master - but if that works great!

bigendiansmalls:
Researched the $top_srcdir issue and found this from a while back.  I'll find a different way to make it a non-relative path that passes an out-of-tree build.  

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=402727

bigendiansmalls:
I believe I've cleaned up and streamlined the above issues.  Let me know.  Cheers.

gperciva:
Thanks for all this work, bigendiansmalls!

I have a number of issues to mention, but before I do, I wanted to offer to tackle them myself.  If you're eager and keen, then great!  But if you feel like I'm nitpicking (or simply wouldn't mind me contributing as well), then I'm quite willing to work any (or all!) of these.

1) it's getting a bit difficult to track which changes to which files are the final version.  If you're comfortable with `git rebase`, would you mind combining/reorganizing these commits based on overall logic (instead of time)?  For example, I'm imagining one commit which adds `test_scrypt` to the automake, one commit which changes the README, and assorted commits for other issue.

1b) For example, the check for `openssl/aes.h` looks very valuable.  If that was on a commit by itself, we could examine and pull that commit by itself, while we were still discussing/reworking other parts of these changes.  (also, when checking the commit history later for whatever reason, a commit with the subject "configure: check for openssl/aes.h" is very easy to understand, whereas if that change is buried in a much larger "Fixed test functionality during make" then it might be missed)

[if you would like me to extract a few bits and pieces like that, I'm happy to do so, and of course I would retain your name on those commits]

1c) another reason for suggesting this is that when I do this type of "commit refactoring", I often discover changes I've made during development which aren't actually necessary.  Splitting up the patches makes it easier for everybody to check for the highest quality.

2) could you convince me that `%reldir%` is necessary?  I just tried removing `%reldir%/` from `Makefile.am`, and it seems to work in an out-of-tree build.

3) the README instructions don't work for running `test_scrypt.sh` in an out-of-tree build.  (I think some more magic may be required to have a test system that works with in-tree and out-of-tree builds... hmm, my first thought is to have a makefile target, which passes any necessary information to `test_scrypt.sh` on the command-line.  (things like the location of the `test_scrypt` binary and the `test_scrypt.good` file)

4) with an in-tree build, I see this:
```
td@gin: ~/src/scrypt/test (bigendiansmalls-update-test)
$ ./test_scrypt.sh 
./test_scrypt.sh: 18: [: true: unexpected operator
TESTS PASSED!
```

Apparently I'm running dash?  (ubuntu 14.04)
```
td@gin: ~/src/scrypt/test (bigendiansmalls-update-test)
$ ls -l /bin/sh
lrwxrwxrwx 1 root root 4 Apr  5  2015 /bin/sh -> dash
```

unfortunately I know little about shells and portability, so I can't judge whether this is a dash bug or something that could be fixed in the script.

bigendiansmalls:
Hi Graham - thanks!

Yes- I would be happy to make all of the above changes.  Very comfortable with rebase and breaking them out makes a good deal of sense.    

Also, I'll try to make the sh script more agnostic to the running shell without overcomplicating it.  Ideally it would run as part of the finished Make on its own, thus validating the installation, as you state.  I think this can be done reasonably easily.

Let me redo the commits and create separate requests for them, if that works for you.  We can leave this open and close it out once the others are done.  Does that work?
Thinking:

1) aes change in configure.  Simple and stands alone.
2) Build for test_scrypt binary.
3) Build for test_scrypt.sh shell script & test inclusion in Make
4) Updated Readme with the info about how to do the install / tests as they'll be done at that point.

I could not get an in-tree build to work without %reldir%  Let me do some other testing and find the edge cases.

I'll also keep the change from your other pull out of this, so as you state it can be done easily and granularly.

Happy to help - great product all around.

Cheers.

gperciva:
Splitting it into those 4 PRs looks good!  I can run some more tests about `%reldir%` if you want.  At the moment I've got a pretty-much stock Ubuntu 14.04 installation, but I have a few virtual machines laying around so I could test on some other distros (and freebsd) if you think that would help.

bigendiansmalls:
Sure it can't hurt, I run just about 1 of every flavor there is so I can try a multitude of in and out of band tests too.  Let me know if you find anything.

cperciva:
We can close this PR since it's coming back in pieces, right?

bigendiansmalls:
yes. that should be just fine.

PR #23, opened 2015-12-03
Revert "autoconf compiler warnings"

gperciva:
This reverts commit 5d232cf882e985fbb4d037cb3ad8cae8f986640c.

The AX_FLAGS_WARN_ALL macro isn't in Ubuntu 14.04, it isn't included in
scrypt, and since it's GPLv3, I'm not certain we want to distribute it.
(it's not source code, but is "part of the build system" sufficiently separate
from the "derivative work" question?  I'd need to spend a while looking into
this before I'd want to form an opinion one way or the other)

I suggest we disable this for now (to avoid people seeing

    ../configure: line 5050: AX_CFLAGS_WARN_ALL: command not found

when attempting to configure scrypt), and potentially re-enable it if/when the
legal question (and the technical question of how to include it, if we go that
route) is resolved.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -48,15 +48,16 @@ AC_CHECK_FUNCS([mmap])
 
 AC_SYS_LARGEFILE
 
+# Check whether the user has requested to disable compiler warnings
 AC_MSG_CHECKING([compiler_warnings])
 AC_ARG_ENABLE(compiler_warnings,
- AS_HELP_STRING([--disable-compiler-warnings],
-  [Do not request compiler warnings. @<:@kefault=enabled@:>@]),
- [ac_compiler_warnings=$enableval],
- [ac_compiler_warnings=yes])
+   AS_HELP_STRING([--disable-compiler-warnings],
+       [Do not request compiler warnings. @<:@kefault=enabled@:>@]),
+   [ac_compiler_warnings=$enableval],
+   [ac_compiler_warnings=yes])
 AC_MSG_RESULT([${ac_compiler_warnings}])
 AS_IF([test x${ac_compiler_warnings} = xyes],
- [AX_CFLAGS_WARN_ALL])
+   [AX_CFLAGS_WARN_ALL])
 
 AC_CONFIG_HEADERS([config.h])
 AC_CONFIG_FILES([Makefile])
--- a/configure.ac
+++ b/configure.ac
@@ -48,7 +48,15 @@ AC_CHECK_FUNCS([mmap])
 
 AC_SYS_LARGEFILE
 
-AX_CFLAGS_WARN_ALL
+AC_MSG_CHECKING([compiler_warnings])
+AC_ARG_ENABLE(compiler_warnings,
+ AS_HELP_STRING([--disable-compiler-warnings],
+  [Do not request compiler warnings. @<:@kefault=enabled@:>@]),
+ [ac_compiler_warnings=$enableval],
+ [ac_compiler_warnings=yes])
+AC_MSG_RESULT([${ac_compiler_warnings}])
+AS_IF([test x${ac_compiler_warnings} = xyes],
+ [AX_CFLAGS_WARN_ALL])
 
 AC_CONFIG_HEADERS([config.h])
 AC_CONFIG_FILES([Makefile])


bigendiansmalls:
@gperciva I'd noticed this too and removed the line from my most recent (nearly pushed) PR.  Do you want me to leave it in / remove it / don't care?

gperciva:
I'd rather get this tiny change accepted, then you can do a rebase on that version of `origin/master`.  This way, if/when we revisit using `AX_CFLAGS_WARN_ALL`, it will be easier to track what happened when, and also easier for git to avoid merge conflicts.

cperciva:
This bit came from @barak as part of #8 ... I wonder if he can comment on this?  I had the impression that it was a standard option which Debian uses in all of the packages they build, but maybe not...

gperciva:
It's not included in debian 8 (current stable); I just checked by trying to configure scrypt, and got the same warning in `configure`.

If the macro in included in the developer's version of autotools, it appears to add `-Wall -ansi -pedantic` (if using gcc; apparently it doesn't add anything for clang? it's not mentioned in the docs, at least).  http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_cflags_warn_all.m4

Somewhat underwhelming, given the annoying (although admittedly harmless) `command not found` message that gets printed.

barak:
AX_CFLAGS_WARN_ALL is in the Debian (and presumably Ubuntu) package autoconf-archive.

gperciva:
Thanks for the update, @barak!  I can confirm that autoconf-archvie exists on Ubuntu as an optional universe/devel package.  It's also available as a freebsd port, http://portsmon.freebsd.org/portoverview.py?portname=autoconf-archive

Quick refresher for @cperciva: Ubuntu has four "levels" of repositories.  "Main" (officially supported), "restricted" (officially supported but non-free; probably mainly stuff like nvidia drivers but don't quote me on this), "universe" (community-maintained free), and "multiverse" (community-maintained non-free).  I guess that main is like freebsd base, while universe is like the ports.

If autoconf-archive is installed, then when we prepare a tarball with `release-tools/mktarball.sh`, we get an extra `-Wall` added to `CFLAGS`.  (I tried autoconf and running configure with and without `AX_CFLAGS_WARN_ALL` then diff'd the resulting `Makefile`s)

I still recommend that we do not include `AX_CFLAGS_WARN_ALL` in the main autoconf script.  If the scrypt source code still only existed on @cperciva's hard drive (and backups) and people only saw the official scrypt releases, then I don't see any problem with `AX_CFLAGS_WARN_ALL` since it wouldn't appear in user-visible files.  But now that it's on github, I think that enough people will be forking the repo and compiling the github version, and not all of those people will have `autoconf-archive` installed.

barak:
The logic in favour is that with security-sensitive code like this, you *really* want to see all the warnings, and the appropriate flags for all-reasonable-warnings will vary from compiler to compiler (gcc, clang, icc, etc).

cperciva:
Is there any straightforward way to make this a configure option?  `--cflags-warn-all`, say?  That would make it available for people who want it (I certainly see the why it's useful) without adding the unnecessary build dependency for most users.

barak:
The macro is automatically cached in ```aclocal.m4``` when ```autoreconf``` is run, and that file would be included in the ```make dist``` distribution tarball. (In fact, that's the whole point of ```aclocal.m4```.) So end users receiving that tarball would be able to regenerate ```configure``` and friends. That behaviour was frozen years ago, before people wanted to regenerate everything possible to avoid version skew bugs and the like, and before people stared to use git tags to generate distribution tarballs and wanted to avoid committing generated files like that to their repos.

Another option would be to cache the code in the directory where local m4 macros live.
````sh
$ mkdir -p m4
$ cp /usr/share/aclocal/ax_cflags_warn_all.m4 m4/
$ git add m4/ax_cflags_warn_all.m4
````
This is recommended in the autotools documentation. Aside from the wasted space, its only disadvantage is the need to refresh that cached copy when a new version of ```ax_cflags_warn_all.m4``` comes out.

barak:
I've created a branch ```guard-wall``` in my fork which adds a ```./configure --disable-compiler-warnings``` option to skip the macro in question, commit 7be2654.

gperciva:
Thanks, @barak!  I've updated this PR to include your `./configure --disable-compiler-warnings`, along with some minor changes to match the existing style of @cperciva's `configure.ac`.

Colin, should I add apply this solution to Tarsnap's other projects?

cperciva:
Thanks, this looks good now.

@barak I'm deliberately not distributing `aclocal.m4`, and not adding that script to `/m4/`, because I don't want to distribute any GPLv3 files.  The configure script itself is fine, of course, due to the license exception.

@gperciva The only other place I use autotools is in the tarsnap client; but yes, please go ahead and apply this there if we didn't inherit anything equivalent from libarchive.

PR #24, opened 2015-12-08
Updated configure.ac to check for openssl/aes.h

bigendiansmalls:
Forced configure to fail if it cannot find openssl/aes.h on the system.
Without this, the make fails.

Prior to this patch, make can fail (after a successful configure) with an error such as this: 
image

Post patch configure fails like this:
image

A successful configure post patch:
image

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -36,6 +36,11 @@ AC_CHECK_MEMBERS([struct sysinfo.totalram, struct sysinfo.mem_unit], [], [],
 # to work.
 AC_CHECK_HEADERS([sys/param.h])
 
+# Check if we have <openssl/aes.h>, to fix the make failing when configure 
+# passes.
+AC_CHECK_HEADERS([openssl/aes.h],,
+  [AC_MSG_ERROR([Unable to find the openssl/aes.h header])])
+
 # Check for <sys/sysctl.h>.  If it exists and it defines HW_USERMEM
 # and/or HW_MEMSIZE, we'll try using those as memory limits.
 AC_CHECK_HEADERS([sys/sysctl.h])


bigendiansmalls:
I tested it on these 3 current kernel versions
Debian 4.0.4
15.0.0 Darwin Kernel Version 15.0.0
4.2.5-1-ARCH

gperciva:
Thanks, bigendiansmalls!

I think the patch can be reduced to 3 lines:
```
# Check if we have <openssl/aes.h>, to fix the make failing when configure passes.
AC_CHECK_HEADERS([openssl/aes.h],,
    [AC_MSG_ERROR([Unable to find the openssl/aes.h headers])])
```

NOTE: there are two commas `,,` at the end of the `AC_CHECK_HEADERS` line.  This is important, because the first comma means "what to do if I found the header", while the second comma means "what to do if I did not find the header".

Could you check that on your system, and if it works, update this PR?  It looks like it works on my ubuntu 14.04 system.

Also, I'm quite content to see the screenshots you provided, but if you find it easier to just copy&paste text, you can do that by writing three backticks in a row, paste the material, then add three backticks afterwards.

\`\`\`
      <-paste material here
\`\`\`

Cheers,
- Graham

bigendiansmalls:
Cool I'll make those changes.  And yes, I'm familiar with Github's markdown - but sometimes a picture is just more fun.  :)

bigendiansmalls:
Fixed as requested!

gperciva:
Thanks!  Two nitpicks:
- please stay within 80 characters (so put "passes." onto a second comment line, like the other comments in that file do)
- please rebase these two (or three, after the above change) into a single commit.  That keeps the history cleaner.  :)

bigendiansmalls:
No problem.   Done & done.

gperciva:
LGTM.  @cperciva, time to pull it?

Issue #1, opened 2015-05-22
Opening files in binary mode.


GoogleCodeExporter:
```
I found this bug when I modified the original code to run on windows.
I have modified fopens to use binary mode in main.c .
Modified main.c is in the attachment.
```

Original issue reported on code.google.com by `doma...@gmail.com` on 4 Jan 2014 at 12:11

Attachments:
* [main.c](https://storage.googleapis.com/google-code-attachments/scrypt/issue-1/comment-0/main.c)

cperciva:
Fixed by 665d284d.

PR #26, opened 2015-12-10
Fixed makefile build for test/test_scrypt

bigendiansmalls:
This pull fixes the build of the test/test_scrypt executable.

* It removes the Makefile from the test directory
* Centralizes build in one makefile
* renames the test/main.c to test/test_scrypt.c

Tested on a  Linux (Kali & Ubuntu) and Mac (El Capitan).
* Worked for both in and out of tree builds.

Cheers.


gperciva:
Thanks for this update, bigendiansmalls!  I gave two nitpicks on the patch itself, but it generally looks really good.  :)

bigendiansmalls:
Willdo.

bigendiansmalls:
@gperciva - Not what I should do with this:

"We don't actually need to include test/test_scrypt.c in EXTRA_DIST: since it's mentioned as one of the build targets, it's included in make dist."

In the Makefile.am - the original had:
`test/main.c`
I updated it to 
`test/test_scrypt.c`

Perhaps leave this for now (as it's just a rename / technical fix to match what was there).  Then in a future PR could test out the distribution (via the script you mention) and see if it's needed at all?

Or would you want me as part of this to just remove the "test/..." lines from EXTRA_DIST?   Judging by your previous comments, I suspect the former - but just wanted to be clear.   I'll sort out the tabs (vim settings wrong :(  ) and rebase/push that up, awaiting your response to question #1.

Cheers.

bigendiansmalls:
Updated.

gperciva:
Hi @bigendiansmalls,

Sorry for the delay about the `EXTRA_DIST` question.  Apparently that's only used for the debian pacakge; I've started looking at https://github.com/barak/scrypt/tree/debian to see if there's stuff I should be merging upstream.

This doesn't meant that your PR should be delayed, though.  It all looks good to me now, so I'm asking @cperciva to merge it now.

cperciva:
Hmm, this may be a silly question, but test_scrypt doesn't seem to actually be building...?

gperciva:
For forgot to run

    autoreconf -i 

before `configure` and `make`.

cperciva:
Nope, did that.  Definitely not building...

cperciva:
Oops, I was just running `make` and forgot that there was a BSDmakefile.  Running `gmake` builds it.

PR #27, opened 2015-12-21
Remove unnecessary files from test_scrypt compile

gperciva:
This follows the original `SRCS` in the previous `test/Makefile` exactly.  As a
result, it does not have any `*_aesni*` files, which may or may not be desirable.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -76,21 +76,12 @@ EXTRA_DIST = \
 	test/test_scrypt.good					\
 	test/test_scrypt.sh
 
-test_test_scrypt_SOURCES =  test/test_scrypt.c			\
-		libcperciva/alg/sha256.c			\
-		libcperciva/cpusupport/cpusupport_x86_aesni.c	\
-		libcperciva/cpusupport/cpusupport_x86_sse2.c	\
-		libcperciva/crypto/crypto_aes.c			\
-		libcperciva/crypto/crypto_aesctr.c		\
-		libcperciva/crypto/crypto_entropy.c		\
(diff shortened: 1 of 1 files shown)


cperciva:
I think what we have here was correct already... `crypto_aes_aesni.c` and `crypto_scrypt_smix_sse2.c` get compiled in via `libcperciva_aesni.a` and `libscrypt_sse2.a` since autoconf is incapable of building different C files with different options if they belong to the same target.

gperciva:
This included `scryptenc.h`, whereas `test_scrypt.c` only includes `crypto_scrypt.h`.  I discovered this because https://github.com/Tarsnap/scrypt/tree/command-line-verbose wouldn't compile with this change until I added `humansize.c` to the target for `test_scrypt`, which was silly since it was never called.

I've updated the file list to only remove `scryptenc.c` and related files.

cperciva:
Oh right, this test code is just for scrypt itself, not for the file encryption bits.  Not doing any encryption means that we don't need `cpusupport_x86_aesni.c`, `libcperciva_aesni.a`, `crypto_aesctr.c`, or `crypto_aes.c`.  Also I'm pretty sure we don't need `crypto_entropy.c` or `entropy.c`.

gperciva:
Updated; it compiles and passes the tests without `*entropy.c`, so I removed those as well.

It also looks like you prefer a tab after a line like

    test_test_scrypt_SOURCES =

so I've tweaked the whitespace accordingly.

gperciva:
Updated the commit message to document what's happening with `crypto_scrypt_smix_sse2.c`

PR #29, opened 2016-01-19
Magic getopt

gperciva:
I also added error-checking for things like

    scrypt enc -M -v in.txt out.txt

(since the getopt treats `-v` as the argument to `-M`)


gperciva:
I tried adding some error-handling code into getopt; please see #30 instead.

PR #30, opened 2016-01-19
Magic getopt error

gperciva:
This is an improved version of #29 with better error-handling.


gperciva:
You can see the old error-handling code in https://github.com/Tarsnap/scrypt/tree/magic-getopt

PR #32, opened 2016-01-27
Add scrypt --version: no "mode" variable

gperciva:


cperciva:
Ok, that idea didn't work.  But looking at the code just now... is there a reason we can't just move the handling of `--version` up to where you have `/* DO NOTHING */`?

gperciva:
For similarity with other Tarsnap projects, I thought `--version` would look better being handled by magic getopt, but I must admit that this new diff looks much better!

cperciva:
Yeah, I had also been thinking of it as something to handle with GETOPT, but this seems like a vast improvement.  Lesson learned: Just because we have a nifty new hammer doesn't mean that everything is a nail...

PR #33, opened 2016-01-28
Add scrypt --version: add a "mode" variable

gperciva:
Alternate version of #32 .  (added as a separate PR in case you want to compare the two methods)


cperciva:
Ick.  No, I was wrong, this is even worse.  Thanks for writing the code so that I could see what it looked like though...

PR #34, opened 2016-03-03
Provide -P option to read password from stdin

ijackson:
This makes it easier to use the command-line scrypt utility in scripts etc.  This is particularly useful because the scrypt utility contains some parameter choosing and storing logic which is not exposed in a cooked way elsewhere.

Signed-off-by: Ian Jackson <ijackson@chiark.greenend.org.uk>

### DIFF
--- a/main.c
+++ b/main.c
@@ -55,6 +55,7 @@ main(int argc, char *argv[])
 {
 	FILE * infile;
 	FILE * outfile;
+	int devtty = 1;
 	int dec = 0;
 	size_t maxmem = 0;
 	double maxmemfrac = 0.5;
@@ -98,6 +99,9 @@ main(int argc, char *argv[])
 		GETOPT_OPT("-v"):
 			verbose = 1;
 			break;
+		GETOPT_OPT("-P"):
+			devtty = 0;
+			break;
 		GETOPT_MISSING_ARG:
 			warn0("Missing argument to %s\n", ch);
 			/* FALLTHROUGH */
@@ -134,7 +138,7 @@ main(int argc, char *argv[])
 
 	/* Prompt for a password. */
 	if (readpass(&passwd, "Please enter passphrase",
-	    dec ? NULL : "Please confirm passphrase", 1))
+	    (dec || !devtty) ? NULL : "Please confirm passphrase", devtty))
 		exit(1);
 
 	/* Encrypt or decrypt. */
--- a/libcperciva/util/readpass.c
+++ b/libcperciva/util/readpass.c
@@ -39,6 +39,22 @@ handle(int sig)
 }
 
 static void
+resetsigs(struct sigaction *savedsa)
+{
+	int i;
+
+	/* Restore old signals. */
+	for (i = 0; i < NSIGS; i++)
+		sigaction(badsigs[i], &savedsa[i], NULL);
+
+	/* If we intercepted a signal, re-issue it. */
+	for (i = 0; i < NSIGS; i++) {
+		if (gotsig[badsigs[i]])
+			raise(badsigs[i]);
+	}
+}
+
+static void
 warnp_read(FILE *readfrom)
 {
 	if (feof(readfrom))
@@ -139,15 +155,7 @@ retry:
 	if (usingtty)
 		tcsetattr(fileno(readfrom), TCSANOW, &term_old);
 
-	/* Restore old signals. */
-	for (i = 0; i < NSIGS; i++)
-		sigaction(badsigs[i], &savedsa[i], NULL);
-
-	/* If we intercepted a signal, re-issue it. */
-	for (i = 0; i < NSIGS; i++) {
-		if (gotsig[badsigs[i]])
-			raise(badsigs[i]);
-	}
+	resetsigs(savedsa);
 
 	/* Close /dev/tty if we opened it. */
 	if (readfrom != stdin)
--- a/libcperciva/util/readpass.c
+++ b/libcperciva/util/readpass.c
@@ -196,6 +196,8 @@ err1:
 	insecure_memzero(passbuf, MAXPASSLEN);
 	insecure_memzero(confpassbuf, MAXPASSLEN);
 
+	resetsigs(savedsa);
+
 	/* Failure! */
 	return (-1);
 }
--- a/libcperciva/util/readpass.c
+++ b/libcperciva/util/readpass.c
@@ -38,6 +38,15 @@ handle(int sig)
 	gotsig[sig] = 1;
 }
 
+static void
+warnp_read(FILE *readfrom)
+{
+	if (feof(readfrom))
+		fprintf(stderr, "EOF reading password\n");
+	else
+		warnp("Cannot read password");
+}
+
 /**
  * readpass(passwd, prompt, confirmprompt, devtty)
  * If ${devtty} is non-zero, read a password from /dev/tty if possible; if
@@ -104,7 +113,7 @@ retry:
 
 	/* Read the password. */
 	if (fgets(passbuf, MAXPASSLEN, readfrom) == NULL) {
-		warnp("Cannot read password");
+		warnp_read(readfrom);
 		goto err3;
 	}
 
@@ -113,7 +122,7 @@ retry:
 		if (usingtty)
 			fprintf(stderr, "%s: ", confirmprompt);
 		if (fgets(confpassbuf, MAXPASSLEN, readfrom) == NULL) {
-			warnp("Cannot read password");
+			warnp_read(readfrom);
 			goto err3;
 		}
 		if (strcmp(passbuf, confpassbuf)) {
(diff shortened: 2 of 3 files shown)


main.c
```diff
@@ -134,7 +138,8 @@ main(int argc, char *argv[])
 
 	/* Prompt for a password. */
 	if (readpass(&passwd, "Please enter passphrase",
-	    dec ? NULL : "Please confirm passphrase", 1))
+	    dec ? NULL : !devtty ? NULL :
```
cperciva: I'm not sure what the point of this is... if we're not using a tty, we won't be printing a prompt anyway, I think?

ijackson: Colin Percival writes ("Re: [scrypt] Provide -P option to read password from stdin (#34)"):
> > @@ -134,7 +138,8 @@ main(int argc, char *argv[])
> >  	/* Prompt for a password. */
> >  	if (readpass(&passwd, "Please enter passphrase",
> > -	    dec ? NULL : "Please confirm passphrase", 1))
> > +	    dec ? NULL : !devtty ? NULL :
> 
> I'm not sure what the point of this is... if we're not using a tty, we won't be printing a prompt anyway, I think?

I may have misunderstood the code.  Feel free to drop that aspect of
it.  Would you like me to retest without that change ?

Thanks,
Ian.

cperciva:
Yes, please test without that... it's entirely possible that I've misunderstood the code too.  I wrote this a long time ago. ;-)

main.c
gperciva: It looks like we want `confirmprompt` (the 3rd argument of `readwd()`) to be `NULL`.  From `readpass.h`:

>  If ${confirmprompt} is non-NULL, read a second password (prompting if a terminal is being used) and repeat until the user enters the same password twice.

That's born out in the code:
```
        /* Confirm the password if necessary. */
        if (confirmprompt != NULL) {
                if (usingtty)
                        fprintf(stderr, "%s: ", confirmprompt);
                if (fgets(confpassbuf, MAXPASSLEN, readfrom) == NULL) {
                        warnp("Cannot read password");
                        goto err3;
                }
                if (strcmp(passbuf, confpassbuf)) {
                        fprintf(stderr,
                            "Passwords mismatch, please try again\n");
                        goto retry;
                }
        }
```

I don't see the point of sending a password over `stdin` twice, so I think we should keep that `!devtty ? NULL`.

ijackson: Graham Percival writes ("Re: [scrypt] Provide -P option to read password from stdin (#34)"):
> > @@ -134,7 +138,8 @@ main(int argc, char *argv[])
> >  
> >  	/* Prompt for a password. */
> >  	if (readpass(&passwd, "Please enter passphrase",
> > -	    dec ? NULL : "Please confirm passphrase", 1))
> > +	    dec ? NULL : !devtty ? NULL :
> 
> It looks like we want `confirmprompt` (the 3rd argument of `readwd()`) to be `NULL`.  From `readpass.h`:

Right, I think that's what I was thinking.

> >  If ${confirmprompt} is non-NULL, read a second password (prompting if a terminal is being used) and repeat until the user enters the same password twice.
> 
> That's born out in the code:
> ```
>         /* Confirm the password if necessary. */
>         if (confirmprompt != NULL) {
>                 if (usingtty)
>                         fprintf(stderr, "%s: ", confirmprompt);
>                 if (fgets(confpassbuf, MAXPASSLEN, readfrom) == NULL) {
>                         warnp("Cannot read password");
>                         goto err3;
>                 }
>                 if (strcmp(passbuf, confpassbuf)) {
>                         fprintf(stderr,
>                             "Passwords mismatch, please try again\n");
>                         goto retry;
>                 }
>         }
> ```
> 
> I don't see the point of sending a password over `stdin` twice, so I think we should keep that `!devtty ? NULL`.

Right.  Certainly scrypt(1)'s -P option shouldn't read the pw twice.

Arguably that readpass does so in the non-tty case is undesirable but
I guess you don't want to change that.

Thanks,
Ian.

cperciva: Ok, this makes sense.  Sorry about the delay -- I wanted to check how I was using `readpass` in other code in case I wanted to change the non-tty behaviour, but I think it's best to leave that as reading twice even if we *think* we're not reading from a tty in case there's weirdness going on and we don't detect it correctly.

One small request, can you spell `dec ? NULL : !devtty ? NULL : ...` as `(dec || !devtty) ? NULL : ...` instead?  Seems both simpler and easier to understand.

cperciva:
Aside from the `?:?:` bit I mentioned above, the only problem I see here is that there is no documentation.  Do you think you could update `scrypt.1`?  (If you're not comfortable with editing man pages, I'll merge without it and get @gperciva to update the documentation.)

ijackson:
Colin Percival writes ("Re: [scrypt] Provide -P option to read password from stdin (#34)"):
> Aside from the `?:?:` bit I mentioned above, the only problem I see here is that there is no documentation.  Do you think you could update `scrypt.1`?  (If you're not comfortable with editing man pages, I'll merge without it and get @gperciva to update the documentation.)

Sure.  I'll do that and update the pull request.

Ian.

ijackson:
Colin Percival writes ("Re: [scrypt] Provide -P option to read password from stdin (#34)"):
> Aside from the `?:?:` bit I mentioned above, the only problem I see here is that there is no documentation.  Do you think you could update `scrypt.1`?  (If you're not comfortable with editing man pages, I'll merge without it and get @gperciva to update the documentation.)

I have now done this and updated the pull request branch.

You'll see that I added 3 more error handling patches.  If you don't
like them, I won't be offended if you don't take them.  In particular,
  Always clean up on success path even if signaled
is quite arguably marginal.

These commits are on top of the branch so you can easily drop them.
(Weirdly, looking at the pull request page in github shows the commits
in an odd order.)  Or just let me know and I'll re-push with them
omitted.

Regards,
Ian.

cperciva:
Could you submit the `readpass` changes as a separate PR to libcperciva?  I don't want to lose them completely (although I'm skeptical about dda36e2a) but they don't belong as part of this PR.

gperciva:
Hi @ijackson, would you mind if I helped out with this functionality?  I'm adding some more tests to scrypt (encrypt a file, check that it differs from the original, try to decrypt it with the wrong password, etc.), and the `-P` option would be very useful when I'm writing those scripts.  :)

My first thought is to make a new PR with only the `-P` and man-page update, leaving this PR for the additional `readpass` changes.  I could also take a stab at addressing Colin's concerns.

ijackson:
Graham Percival writes ("Re: [scrypt] Provide -P option to read password from stdin (#34)"):
> Hi @ijackson, would you mind if I helped out with this functionality?  I'm adding some more tests to scrypt (encrypt a file, check that it differs from the original, try to decrypt it with the wrong password, etc.), and the `-P` option would be very useful when I'm writing those scripts.  :)
> 
> My first thought is to make a new PR with only the `-P` and man-page update, leaving this PR for the additional `readpass` changes.  I could also take a stab at addressing Colin's concerns.

Sure, please, go ahead.

One other thing I noticed is that the exit status from scrypt(1) could
be improved.

ATM it calls mainly exit(1).  I would like it to call exit(rc) at the
bottom of main, and change the others to exit(-1).  What do you
(plural) think ?

Ian.
I

gperciva:
Calling `exit (1)` is the typical @cperciva way of exiting a program, so I'm not optimistic about changing it.  In particular, if we start returning different exit codes, we need to commit to always using them in the same way in the future.  I see why this could be useful to people running scrypt in a script, but I would be surprised if @cperciva wanted to commit to specific exit codes at this time.

gperciva:
Closing this because development continues in #41 .

ijackson:
Graham Percival writes ("Re: [Tarsnap/scrypt] Provide -P option to read password from stdin (#34)"):
> Calling exit (1) is the typical @cperciva way of exiting a program, so I'm not
> optimistic about changing it. In particular, if we start returning different
> exit codes, we need to commit to always using them in the same way in the
> future. I see why this could be useful to people running scrypt in a script,
> but I would be surprised if @cperciva wanted to commit to specific exit codes
> at this time.

Note that these changes are to the exit status of scrypt only.

And indeed, it is important to anyone running scrypt in a script or as
a subprocess.

Running scrypt as a subprocess is currently the only way to access the
autoscsaling functionality, right now.  I considered refactoring
things to provide that in the library, but since I was wanting to run
scrypt from Perl, I wanted a subprocess interface anyway.

The alternative to a defined exit status is for the caller to parse
the error message, which is very bad for several reasons.

Note that it is still a good idea to have a fallback exit status which
means `something hideous went wrong'.  Typically 127 is good for that,
but 1 will do if it isn't used for anything else.

Libraries should never exit with 1 (if they exit at all).  This is
because otherwise it becomes impossible to use such a library safely
in a program which is expected to provide a defined exit status (eg,
something which is expected to replace an implementation of diff, or a
mail delivery program).

If a library is defined to (or needs to) exit the whole program, it
should call exit(-1) or exit(127).

Ian.

cperciva:
I don't know what you're talking about re exit status... none of the library code calls `exit` at all...?

PR #35, opened 2016-03-10
build: add "make test"

gperciva:
This is functionality split off from @bigendiansmalls' PR:
https://github.com/Tarsnap/scrypt/pull/22

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -90,3 +90,11 @@ test_test_scrypt_SOURCES =	test/test_scrypt.c			\
 			lib/crypto/crypto_scrypt_smix.c			\
 			cpusupport-config.h
 test_test_scrypt_LDADD=	libscrypt_sse2.a
+
+.PHONY: test
+# we can't only build "scrypt test/test_scrypt" because that won't build
+# the BUILT_SOURCES.
+test: all
+	$(top_srcdir)/test/test_scrypt.sh \
+		test/test_scrypt \
+		$(top_srcdir)/test/test_scrypt.good
--- a/README.md
+++ b/README.md
@@ -98,6 +98,14 @@ int crypto_scrypt(const uint8_t *, size_t, const uint8_t *, size_t, uint64_t,
 ```
 
 
+Testing
+-------
+
+A small test suite can be run with:
+
+    make test
+
+
 Mailing list
 ------------
 
--- a/test/test_scrypt.sh
+++ b/test/test_scrypt.sh
@@ -1,7 +1,17 @@
 #!/bin/sh
 
-./test_scrypt > test_scrypt.log
-if cmp -s test_scrypt.log test_scrypt.good; then
+test_scrypt_binary=$1
+test_scrypt_good=$2
+
+if [ -z $test_scrypt_binary ] || [ -z $test_scrypt_good ]; then
+	echo -n "Test binary or good file not given: "
+	echo "Using default values for in-source-tree build."
+	test_scrypt_binary="./test_scrypt"
+	test_scrypt_good="./test_scrypt.good"
+fi
+
+$test_scrypt_binary > test_scrypt.log
+if cmp -s test_scrypt.log $test_scrypt_good; then
 	rm test_scrypt.log
 	echo "PASSED!"
 else


gperciva:
This does not include documentation yet; I was thinking of adding that in a separate PR, but we could merge it with this one instead if you want.

gperciva:
This now includes (brief) documentation about running the test suite.

bigendiansmalls:
FYI - I tested on osx (newest), linux 4.3 i386/64 - all worked well.   One thought, would you want to say "Once compiled, scrypt can be tested with:" vs "The compile can be tested with"  Pedantic, I know, but you can't just do make test, without running make first.  Just a thought.

gperciva:
@bigendiansmalls : I did consider that, so the makefile contains:

    test: scrypt test/test_scrypt

so it *should* work even without running `make`.  And if you run `make && make clean && make test` then it works.  However, I've just discovered that if you go directly to `make test`, it doesn't build `cpusupport.h`.  I recently worked on this in the Tarsnap client repo, so I'll look at getting that improved `cpusupport.h` handling into scrypt as well.

gperciva:
ah, also, I just realized that I didn't update the `BSDmakefile`.  I guess I'd better do that too.  Why do we have two separate build systems for a project which is 3318 lines of code?  :(

gperciva:
Updated to fix `cpusupport.h` issue in a `make test` after `configure`.  `BSDmakefile` has not been changed; is it really necessary to keep it?

Makefile.am
```diff
@@ -90,3 +90,11 @@ test_test_scrypt_SOURCES =	test/test_scrypt.c			\
 			lib/crypto/crypto_scrypt_smix.c			\
 			cpusupport-config.h
 test_test_scrypt_LDADD=	libscrypt_sse2.a
+
+.PHONY: test
+# we can't only build "scrypt test/test_scrypt" because that won't build
+# the BUILT_SOURCES.
```
cperciva: Why does this matter?

Makefile.am
```diff
@@ -90,3 +90,11 @@ test_test_scrypt_SOURCES =	test/test_scrypt.c			\
...
+test: all
+	$(top_srcdir)/test/test_scrypt.sh \
```
cperciva: What's wrong with `(cd test && sh test_scrypt.sh)` ?

cperciva:
If I'm reading some of the comments here correctly, you want `make test` to also test that the compile succeeds?  Isn't that what running `make all` is for?

cperciva:
BTW, if you want to expand testing, you might want to make `test_scrypt.sh` use scrypt to encrypt and decrypt a file...

Makefile.am
gperciva: If we're doing an out-of-src-tree build (which IMO is the only proper way to build stuff -- to clean up, just delete your `my-build/` directory!), then there does not exist any `test/test_scrypt.sh` in the build directory.

Makefile.am
gperciva: If we don't have `BUILT_SOURCES`, then the compile will fail because it doesn't have `cpusupport.h`.  Remember, the way that autotools works in scrypt at the moment is that `make all` works fine, but (from an empty builddir), `make scrypt` will **fail**, because it doesn't build `cupsupport.h`.

scrypt depends on having proper crockery, after all!

gperciva:
Good point on expanding `test_scrypt.sh`; I'll do that after this PR.

As for "testing the compile"... in the first version of this PR, if I did this:
```
mkdir build/
cd build/
../configure
make test
```

it would fail because scrypt hasn't been compiled.  But because of the (new in the second version) dependency `test: all`, running `make test` by itself will ensure that it has compiled `scrypt`.  It's not likely that many people would try testing it without compiling first, but since it was trivial to fix that potential flaw, there's no reason not to support this scenario.

Makefile.am
gperciva: BTW, some (many?) packaging tools only use out-of-tree builds, so making sure we support this isn't just pandering to my OCD.

cperciva: Ick, I hate out-of-tree builds.  Ok, you win.

Makefile.am
cperciva: `make scrypt` seems to work fine for me?  And `cpusupport-config.h` (which I assume is the file you meant) is listed in `scrypt_SOURCES` so unless autotools is even more broken than I thought it should be fine?

cperciva: Wait, never mind, I already had `cpusupport-config.h` built.  Why is it not being built as a dependency when it's listed in `_SOURCES`?

cperciva: Never mind, I think I understand now.  autotools is just being stupid and not realizing what it means for a `.h` file to be a source file.

test/test_scrypt.sh
```diff
@@ -1,7 +1,17 @@
 #!/bin/sh
 
-./test_scrypt > test_scrypt.log
-if cmp -s test_scrypt.log test_scrypt.good; then
+test_scrypt_binary=$1
+test_scrypt_good=$2
+
+if [ -z $test_scrypt_binary ] || [ -z $test_scrypt_good ]; then
+	echo -n "Test binary or good file not given: "
+	echo "Using default values for in-source build."
```
cperciva: s/in-source/in-tree/ ?

README.md
```diff
@@ -98,6 +98,14 @@ int crypto_scrypt(const uint8_t *, size_t, const uint8_t *, size_t, uint64_t,
 ```
 
 
+Testing
+-------
+
+The compile can be tested with
```
cperciva: This tests more than just the compile, yes?

Makefile.am
gperciva: On Sun, Mar 13, 2016 at 01:10:07PM -0700, Colin Percival wrote:
>    Never mind, I think I understand now. autotools is just being stupid and
>    not realizing what it means for a .h file to be a source file.

Yeah.  :(

One attractive option (to me) is to rename it to cpusupport-config.c.
According to the automake manual, this should work much better because it
already knows about .c files being dependencies.  That way we don't need to
mess with BUILT_SOURCES at all.

README.md
gperciva: I was using the grammar "The (validity of the) compile can be tested with", but ok, that's a somewhat obscure form.  :)

gperciva:
PR updated with minor text changes.  (don't merge this version)

cperciva:
Looks good.

gperciva:
updated to rebase history.

PR #38, opened 2016-03-17
Cppflags

gperciva:
This was adapted from @bgilbert's https://github.com/Tarsnap/tarsnap/pull/147.  `scrypt` already used `AM_CPPFLAGS` for the included directories, so I didn't need to update those.


gperciva:
BTW, I checked that it works on ubuntu 14.04  and freebsd 10.2.

PR #40, opened 2016-03-30
Add test for encrypting and decrypting a file

gperciva:


cperciva:
How about adding an encrypted file and confirming that it decrypts correctly?  With the tests as written, the AES code could be replaced by rot13 and they would still pass...

Also, can you hook this into the Travis-CI run?

gperciva:
Updated PR, but not for merging because the variable names could be improved.

Rather, I'm testing the travis-ci.  It looks like it's already working, since it runs `make test` by default, and the "details" show the tests running.

gperciva:
Updated PR, good for merging as far as I'm concerned.

gperciva:
Sorry, ignore that, apparently sh doesn't do boolean logic the way I thought it did.  Still needs more work.

gperciva:
This looks plausible, and seems to behave.

PR #41, opened 2016-03-30
readpass handle EOF and resetsigs

gperciva:
This is a revised version of the readpass changes from #34 .

I'm not certain how we want to handle commit authorship.  It's not important to me, so I propose that once the code is accepted, @ijackson rebases the branch (using the "rebase fix" to combine all my commits into his, and adjusting the commit message if necessary to match the new code).

### DIFF
--- a/libcperciva/util/readpass.c
+++ b/libcperciva/util/readpass.c
@@ -38,6 +38,23 @@ handle(int sig)
 	gotsig[sig] = 1;
 }
 
+/* Restore old signals and re-issue intercepted signals. */
+static void
+resetsigs(struct sigaction savedsa[NSIGS])
+{
+	size_t i;
+
+	/* Restore old signals. */
+	for (i = 0; i < NSIGS; i++)
+		sigaction(badsigs[i], &savedsa[i], NULL);
+
+	/* If we intercepted a signal, re-issue it. */
+	for (i = 0; i < NSIGS; i++) {
+		if (gotsig[badsigs[i]])
+			raise(badsigs[i]);
+	}
+}
+
 /**
  * readpass(passwd, prompt, confirmprompt, devtty)
  * If ${devtty} is non-zero, read a password from /dev/tty if possible; if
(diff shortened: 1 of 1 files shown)


libcperciva/util/readpass.c
```diff
@@ -170,6 +183,8 @@ readpass(char ** passwd, const char * prompt,
 	/* Reset terminal settings if necessary. */
 	if (usingtty)
 		tcsetattr(fileno(readfrom), TCSAFLUSH, &term_old);
+	resetsigs(savedsa);
+
```
cperciva: Shouldn't this be at `err2`?

libcperciva/util/readpass.c
```diff
@@ -38,10 +38,11 @@ handle(int sig)
 	gotsig[sig] = 1;
 }
 
+/* Restore old signals and re-issue intercepted signals. */
 static void
 resetsigs(struct sigaction savedsa[NSIGS])
 {
-	int i;
+	size_t i;
```
gperciva: Not 100% certain you want this, but it seems to match the rest of the code better.

gperciva:
updated PR; yes, `resetsigs()` belonged in `err2`.

gperciva:
This PR contains all the code from #34 , with the exception of https://github.com/ijackson/scrypt/commit/dda36e2a280faab52a288d5b48b32582063a855d "readpass: Always clean up on success path even if signaled", of which @cperciva said:
> I'm not convinced. We needed to capture signals so that we could ensure that we reset the terminal state if we were killed; but if the caller sets up signals to siglongjmp out of readpass then they get what they deserve.

I'm going to close the old PR so that we don't have two PRs for the same code.

gperciva:
Updated PR with rebased and minor rewritten commit messages, with permission from @ijackson.

cperciva:
Thanks!  Sorry it took so long for me to merge this.

PR #45, opened 2016-04-06
Remove unused header file

jgriffiths:
A minor change, but it may help people using this code outside of the library (one less local change to track if they don't have an sha256.h).


cperciva:
Thanks, that was left behind after the `smix` code got split off from the rest of the scrypt code.

I'm not entirely sure what you mean by "don't have an sha256.h" though, since it's part of the tree...?

jgriffiths:
Ah, I just meant if this code is extracted for integration into someones code, i.e. the whole source distribution isn't used. Thanks!

Issue #48, opened 2016-04-20
integerify broken for sse2


jgriffiths:
Hi,

The sse2 variant in lib/crypto/crypto_scrypt_smix_sse2.c defines integerify() with what looks like a clear typo:
```
        const uint32_t * X = (const void *)((uintptr_t)(B) + (2 * r - 1) * 64);
        return (((uint64_t)(X[13]) << 32) + X[0]);
```
Note the indexing of X[13] which it seems should be X[1] as the intent is to read a u64 from 2 u32's.

I believe the reason this has not been discovered in the wild yet is that the result of this function is always and-ed with (N - 1), and no caller ever passes an N > 2^32 (or if they do, they have not compared the sse2 vs general results).

This also suggests that if callers are willing to limit N beyond its current constraint of 'power of 2 >= 2' to include an upper limit of 2^32, that the return can be reduced to simply X[0] and the shift/add removed.

For any caller desiring interop between 32 and 64 bit platforms, the upper limit of N is in any case (2^32)/128.

cperciva:
Are you sure about that?  It's a long time since I wrote this, but I have a vague memory of the data being stored in a permuted order here...

jgriffiths:
I'm sure that the value read isn't used with N < 2^32.

I guess its possible that the data is permuted, given that 13 is used as one of the constants in the inner loop of salsa20_8. How can I confirm that?

jgriffiths:
@cperciva You are correct, the data is permuted - sorry for the noise.

I've changed my PR to update the documentation for this function and not fetch the upper bits when we know that that N cannot be bigger than2^32. If you don't think the change is worth it, I'll nuke the PR in a day or so.

cperciva:
I was going to suggest adding a printf of the output of Integerify, but dumping the entire blocks and checking that the words are permuted also works.  Thanks for confirming.

PR #49, opened 2016-04-20
Integerify update

jgriffiths:
One comment change and two small optimisations relating to issue #48.


cperciva:
Looks like with `-O2` both clang and gcc inline the function and never load the second word on i386, so there's no need to complicate the code.

cperciva:
The comment you added is still worthwhile though, so feel free to open a new PR for that by itself.

PR #51, opened 2016-04-24
build: remove BSDmakefile, move signing to release-tools/

gperciva:
We're now using autotools for building scrypt.  The "make publish" target has
been moved to release-tools/create-sign-tarball.sh


cperciva:
I'm going to merge this now, but in case I look at this later and am confused as to why it isn't working: I now use `env GNUPGHOME=XXXX gpg` instead of `gpg --secret-keyring $GPGKEYFILE` since the secret-keyring option got broken.

PR #52, opened 2016-05-26
Tools: switch from GPGKEYFILE to GNUPGHOME

gperciva:


gperciva:
Addresses https://github.com/Tarsnap/scrypt/pull/51#issuecomment-221741577

cperciva:
Can we make this require `GNUPGSIGNINGHOME` to be set, and then set `GNUPGHOME` to that value?  I could imagine GNUPGHOME being set in a general user environment; using a different name ensures that it was set deliberately for the signing operation.

gperciva:
Do you want a generic `GNUPGSIGNINGHOME`, or `GNUPG_SCRYPT_HOME` OR `GNUPG_TARSNAP_HOME` ?  Or perhaps taking this as a command-line argument, to make extra certain that it's something set deliberately for this signing operation?

cperciva:
Oh, yes, command-line argument.  Sorry, I was still thinking in Makefile mode where everything is variables.

The `-u $GPGKEYID` can be dropped though; the "signing" gnupg home should only have one key so it should be selected automatically.

gperciva:
Updated PR

PR #54, opened 2016-06-26
Getopt error

gperciva:


main.c
```diff
@@ -79,8 +79,10 @@ main(int argc, char *argv[])
 	} else if (strcmp(argv[1], "--version") == 0) {
 		fprintf(stdout, "scrypt %s\n", PACKAGE_VERSION);
 		exit(0);
-	} else
+	} else {
+		warn0("First argument must be 'enc' or 'dec'.\n");
```
gperciva: Yes, technically the first argument could be `--version`, but I think we don't need to mention that in this warning since `usage()` comes right after it.  (I still think it's worth being explicit, since somebody might not be familiar with the terse `{enc | dec}` format.)

PR #60, opened 2016-07-07
Tests: simplify file locations and directories

gperciva:
We only need to pass in the build directory; everything else can be determined from that and $scriptdir.


cperciva:
LGTM.  Is there anything particular I should look at closely before I merge?

gperciva:
No, I think it's good for the merge.

PR #67, opened 2016-09-06
Fix warnings

gperciva:


gperciva:
Found a few more warnings by testing on a 32-bit system.  Nothing serious.

PR #70, opened 2016-10-20
Trivial fixes

gperciva:


lib/crypto/crypto_scrypt.c
```diff
@@ -28,7 +28,6 @@
  */
 #include "scrypt_platform.h"
 
-#include <sys/types.h>
```
cperciva: This was intentional -- older UNIX systems require `<sys/types.h>` for mmap, even though the latest versions of POSIX say that `<sys/mman.h>` should be the only required header.

gperciva:
Updated PR to keep `sys/types.h`

PR #72, opened 2016-10-20
Explicit cast to (uint64_t) from getrlimit()

gperciva:
- This assumes that the amount of available memory available to the process, or the max size of the data segment, or the resident set, will not exceed 18 exibytes (2^64).

- This also assumes that FreeBSD (which uses a signed int for rlim_t, in violation of the POSIX standard) will not return a negative value.


gperciva:
I considered playing some games with `#if (rlim_t)(-1) == (int)(-1)` to be stricter about the assumptions, but I figured you probably wouldn't want that complexity anyway.

PR #74, opened 2016-10-21
Add -f option to force decryption; no checkparams() test

gperciva:


lib/scryptenc/scryptenc.c
```diff
@@ -240,7 +240,8 @@ scryptenc_setup(uint8_t header[96], uint8_t dk[64],
 static int
 scryptdec_setup(const uint8_t header[96], uint8_t dk[64],
     const uint8_t * passwd, size_t passwdlen,
-    size_t maxmem, double maxmemfrac, double maxtime, int verbose)
+    size_t maxmem, double maxmemfrac, double maxtime, int verbose,
+    int force_resources)
```
cperciva: s/force_resources/force/

lib/scryptenc/scryptenc.c
```diff
@@ -269,10 +270,10 @@ scryptdec_setup(const uint8_t header[96], uint8_t dk[64],
 	/*
 	 * Check whether the provided parameters are valid and whether the
 	 * key derivation function can be computed within the allowed memory
-	 * and CPU time.
+	 * and CPU time, unless the user chose to disable this test.
 	 */
-	if ((rc = checkparams(maxmem, maxmemfrac, maxtime, logN, r, p,
-	    verbose)) != 0)
+	if (!force_resources && (rc = checkparams(maxmem, maxmemfrac, maxtime,
```
cperciva: We need to check if the parameters are valid even if we're not trying to limit CPU or RAM usage.

scrypt.1
```diff
@@ -71,6 +72,8 @@ is given, then
 does not print any prompts, and reads a passphrase from stdin.
 .Sh OPTIONS
 .Bl -tag -width "-m maxmemfrac"
+.It Fl f
+Force the decryption to use any required memory or CPU time.
```
cperciva: Don't like the wording.  Either s/Force/Allow/, or further rewording for clarity: "Attempt decryption even if it is anticipated to require an excessive amount of memory or CPU time."

gperciva: I kind-of like the "-f\nForce" bit, though.  What about "Force the decryption to proceed even if it is anticipated to require an excessive amount of memory or CPU time." ?

cperciva: Sure, that works.

gperciva:
PR updated

lib/scryptenc/scryptenc.c
```diff
@@ -168,9 +169,9 @@ checkparams(size_t maxmem, double maxmemfrac, double maxtime,
 
 	/* Check limits. */
 	N = (uint64_t)(1) << logN;
-	if ((memlimit / N) / r < 128)
+	if (!force && ((memlimit / N) / r < 128))
```
cperciva: Why bother calling memtouse and scryptenc_cpuperf if we're simply going to ignore them?  I'd make lines 155-162 conditional on !force and have memlimit and opslimit default to huge values otherwise.

gperciva:
Updated PR; I think that `checkparams()` looks better if we sanity-check `logN` etc. first, and then do all the `force` stuff together.

gperciva:
Incidentally, I do have a test for the `-f` option, but it depends on the "multiple test values" PR #75, so I'll put it forward later.

lib/scryptenc/scryptenc.h
```diff
@@ -91,13 +91,14 @@ int scryptenc_buf(const uint8_t *, size_t, uint8_t *,
 
 /**
  * scryptdec_buf(inbuf, inbuflen, outbuf, outlen, passwd, passwdlen,
- *     maxmem, maxmemfrac, maxtime, verbose):
+ *     maxmem, maxmemfrac, maxtime, verbose, force_resources):
```
cperciva: Not in sync with .c file

lib/scryptenc/scryptenc.h
```diff
@@ -110,11 +111,12 @@ int scryptenc_file(FILE *, FILE *, const uint8_t *, size_t,
 
 /**
  * scryptdec_file(infile, outfile, passwd, passwdlen,
- *     maxmem, maxmemfrac, maxtime, verbose):
+ *     maxmem, maxmemfrac, maxtime, verbose, force_resources):
```
cperciva: (and here too)

gperciva:
Updated PR

gperciva:
Updated PR again to include the test scenario.

Yes, this "re-uses" the number 04, but PR #77 combines the current 02 and 03, so at the end of the day we'll be left with one scenario per number.  (also, the test suite functions perfectly well regardless of the actual filenames, so there's no technical reason to avoid `04-foo.sh` and `04-bar.sh`)

PR #75, opened 2016-10-24
Test valgrind multiple

gperciva:


tests/01-known-values.sh
```diff
@@ -3,25 +3,22 @@
...
-
-### Check output
-scenario_check() {
-	retval=$1
+	# Run the binary which tests known input/output strings.
+	setup_check_variables
+	(							\
+		$c_valgrind_cmd $bindir/tests/test_scrypt	\
+			1> $s_basename-stdout.txt		\
+	; echo $? > $c_exitfile )
```
cperciva: Why is this happening within a subshell?  And why the weird formatting?

tests/01-known-values.sh
```diff
@@ -3,25 +3,22 @@
...
 
 	# The generated values should match the known good values.
-	if ! cmp -s $out/$basename-stdout.txt $scriptdir/test_scrypt.good; then
-		retval=1
+	setup_check_variables
+	if cmp -s $s_basename-stdout.txt $scriptdir/test_scrypt.good; then
+		echo "0" > $c_exitfile
+	else
+		echo "1" > $c_exitfile
```
cperciva: I'd put the `> $c_exitfile` after the `fi` here.

tests/01-known-values.sh
gperciva: Err, if you click on `[...]`, you'll see the commit message.

"""
Note that the awkward

        ( \
        	cmd... \
        ; echo $? > $c_exitfile )

construct is used because ${cmd} might not be exiting before we want the test suite to advance to the next command.  This likely won't apply to test scenarios in scrypt, but when we apply this code to spipe, being able to grab the return code of a server which began 10 shell commands ago is absolutely necessary.  I thought the test suite (as a whole) would be easier to understand if I used that construct everywhere, instead of only for certain scenarios.
"""

I added the extra tab in order to distinguish the "real command" from the boilerplate, but I'm content to remove that if you prefer it removed.

tests/01-known-values.sh
gperciva: woah, I have no idea that was possible!

tests/01-known-values.sh
cperciva: Ah, I was looking at the overall changes, not at the individual commits.  Stick with normal formatting even if some of it is boilerplate:
```
(
    cmd ...
    echo $? > $c_exitfile
)
```

cperciva: BTW if you want to allow ${cmd} to complete later, you probably want to add a `&` after the `)` ending the subshell?

gperciva:
Updated PR:
- removed the "rearrange tests": I'll do that in a separate PR after this one.
- cleaned up the subshell formatting
- cleaned up the `> ${c_exitfile}` stuff
- added assorted `{}` to `$variable_name` for clarity.
- fixed function doc-strings that I forgot to update.

Not a change, but a clarification: I only need the `&` for spiped servers, so that doesn't appear in this code.

PR #77, opened 2016-10-26
Test revise

gperciva:


tests/shared_test_functions.sh
```diff
@@ -221,3 +226,20 @@ scenario_runner() {
...
 	return "${s_retval}"
 }
+
+## run_scenarios (scenario_filenames):
+# Runs all scenarios matching ${scenario_filenames}.
+run_scenarios() {
+	printf -- "Running tests\n"
+	printf -- "-------------\n"
+	scenario_filenames=$@
+	for scenario in ${scenario_filenames}; do
```
cperciva: This is fine, but a more idiomatic approach would be:

```
while [ $# -gt 0 ]; do
    scenario_runner $1
    retval=$?
    if [ ${retval} -gt 0 ]; ten
        exit ${retval}
    fi
    shift
done

cperciva:
I can merge this now, or you can change the for loop syntax first if you'd prefer; I don't have a strong preference, and the "more idiomatic" version is probably less clear to people who aren't native sh speakers...

gperciva:
I prefer the more `for` loop / more readable version.  I was about to ask to wait a bit because I just "noticed" that the globs didn't sort files by alphabetical / numerical value, but apparently this *is* specified in POSIX.

So please merge it if you're satisfied with it.  :)

PR #79, opened 2016-10-26
automake version info

jungle-boogie:
Perhaps not the best place to put the note, but it's at least now on 'paper'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ for experimental development, use:
 In order to support the `AX_CFLAGS_WARN_ALL` autoconf directive, you will need
 to install the autoconf archive.  On Debian systems, use the
 `autoconf-archive` package; on FreeBSD, use `devel/autoconf-archive`.
-
+`automake` versions 1.12 and higher are needed to build scrypt.
 
 Testing
 -------


gperciva:
Thanks for the PR!  Technically, one does not need automake to build scrypt; it is only needed if building code from git instead of an official release tarball.  I wrote about that, and tweaked the formatting, in PR #80.

jungle-boogie:
Ah, I do see that now.

However, your readme doesn't contain information on "building". I went here instead of looking for a git-building page:
https://github.com/Tarsnap/scrypt/blob/master/README.md

gperciva:
Indeed.  PR #80 changes that page: https://github.com/Tarsnap/scrypt/tree/git-build-automake#building. Does that proposed change clarify matters?  If not, I'd like to change it to remove any confusion!  :)

(note that the README.md file is included in the official tarball releases, so it should be written so that it is applicable to both those releases and building from git)

jungle-boogie:
Hi @gperciva,

yes, your link above is clear to me. 

Thanks for this discussion and scrypt.

Thanks,
sean

cperciva:
I've merged #80 -- this one can be closed now, right?

gperciva:
Right, closing it.

PR #82, opened 2016-11-07
Fix release tarball

gperciva:


cperciva:
I was including the files in /autocrap/ deliberately, in response to a request from the Debian maintainer -- I think it was something about Debian policy and rebuilding autogenerated files?  Anyway, they were there for a reason.

gperciva:
I'll look into the debian policy.  I've updated this PR to only include adding the `tests/` directory to the tarball.

PR #83, opened 2016-11-07
Initialize memlimit and opps even when not checked

gperciva:
We print these variables if --verbose is given, so we can't merely leave them uninitialized if -f is specified.


gperciva:
Sorry, I can't believe I forgot to run my `./check.sh` script (which enables tons of warnings).

PR #84, opened 2016-11-07
Test system scrypt

gperciva:


gperciva:
This tests that the compiled `scrypt` can operate with the system's version of scrypt.  Due to the "decrypt a reference file" test, this is already kind-of covered, but this functionality will be necessary when we revise the `spiped` test suite.

tests/shared_test_functions.sh
```diff
@@ -65,6 +65,25 @@ prepare_directories() {
 	fi
 }
 
+## find_system (cmd, args...):
+# Looks for ${cmd} in the $PATH, and ensure that it supports ${args}.
+find_system() {
+	cmd=$1
+	cmd_with_args=$@
+	# Look for ${cmd}.
+	system_binary=`which ${cmd}`
```
cperciva: `which` is not POSIX.  You want `command -v`.

gperciva:
Updated with `command -v`.

PR #86, opened 2016-11-08
Add COPYRIGHT file with 2-clause BSD

gperciva:


gperciva:
I just grabbed the notice from `main.c`, but I see that the file in libcperciva begins with
```
The code and documentation in this directory ("libcperciva") is distributed
under the following terms:
```
Do you want that included here as well?

cperciva:
Follow the example of spiped.  So have an introductory line like that, and also take the union of the copyrights, not just the one from main.c.

gperciva:
Hmm, the spiped COPYRIGHT file doesn't get recognized by github.  I took a look at their infrastructure, and might be able to improve this (although I'm not optimistic).

At any rate, it's more important for the file to be accurate rather than being recognized by github, so I've done a similar thing here.  Also removed `libcperciva/COPYRIGHT` since spiped doesn't have one there.

PR #88, opened 2016-11-13
Travis-CI: enable strict compile warnings/errors

gperciva:


gperciva:
I'm not certain what you'll think about this.  If we had this in place, it would have warned about the potentially uninitialized variables I added in the first merged version of `-f`.  That said, I'm not certain this is the best way to get the extra CFLAGS into the build.

Obvious other ideas are:
- put all the CFLAGS into `.travis.yml` directly.  That would mean that developers wouldn't have a trivially-easy way to compile with those settings.  I'd find that annoying, so I shied away from this solution.
- add these flags to the main build system: many of them are specific to `clang`, so that wouldn't fly.

cperciva:
I'd prefer to put the flags into .travis.yml directly.  Developers can always copy and paste the command from that file anyway...

PR #90, opened 2016-11-15
Travis-CI: use strict compile warnings/errors

gperciva:
This adds code to compile with clang and gcc, since there will be
some difference in their tests.

For the record, the Ubuntu 16.04 "Trusty" image in Travis-CI uses:
- gcc 4.8
- clang 3.5

so the warnings are tailored to those versions.  If/when Travis-CI
updates that build environment, we may need to add other flags
(for example, -Wno-reserved-id-macro to clang).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,17 @@
 language: C
 sudo: required
 dist: trusty
-script: make && make test USE_VALGRIND=1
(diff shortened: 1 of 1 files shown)


gperciva:
(this is a continuation of PR #88, but github "lost" the connection to this branch when I force-pushed it to trigger another test)

PR #91, opened 2016-11-15
Release tarball

gperciva:


cperciva:
I suppose there's no harm in being a bit more polite about autotools...

PR #93, opened 2016-11-21
Typo fix in comment

gperciva:


gperciva:
I'm so proud!  I wrote a whole script to extract comments, handle punctuation [1], and call aspell on the resulting words... and it found 1 typo!  :)

(hopefully I'll find more in our other projects)

[1] because even though command-line aspell handles punctuation just fine, the version you get with python bindings can't even handle `don't`, let alone stuff like `foo):` or `SHA256`.  :(

gperciva:
I'm putting this as https://github.com/Tarsnap/libcperciva/pull/45 instead.

gperciva:
On second thought, we might as well keep this one around, since there's no `md5` or `sha1` to change in this repo.

PR #94, opened 2016-11-22
Clarify comment

gperciva:


gperciva:
Not certain if you'll like this one.  I found

    Equivalent to init(key, nonce);

confusing because it looked as if there would be a literal `^init` function.  I personally would find it easier to understand if there's an indication that you're referring to a function name *ending* in `init`.  I agree that writing out the full function name is a bit verbose, so I'm suggesting adding a single `_` in front.

cperciva:
Good, but we should have `_free` as well.

gperciva:
I didn't notice the `free` because it didn't have `()`.  Mind if I add them?

PR updated.

PR #96, opened 2017-01-09
Work around OS X sysconf(_SC_PHYS_PAGES) glitch

cperciva:
On some versions of OS X (10.10.5, for one), _SC_PHYS_PAGES is defined
but results in a failure with EINVAL.  Treat this as a successful
indication of sysconf being unhelpful rather than as sysconf failing.

### DIFF
--- a/lib/util/memlimit.c
+++ b/lib/util/memlimit.c
@@ -216,8 +216,11 @@ memlimit_sysconf(size_t * memlimit)
 	/* Read the two limits. */
 	if (((pagesize = sysconf(_SC_PAGE_SIZE)) == -1) ||
 	    ((physpages = sysconf(_SC_PHYS_PAGES)) == -1)) {
-		/* Did an error occur? */
-		if (errno != 0)
+		/*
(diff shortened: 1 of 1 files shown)


lib/util/memlimit.c
```diff
@@ -216,8 +216,11 @@ memlimit_sysconf(size_t * memlimit)
 	/* Read the two limits. */
 	if (((pagesize = sysconf(_SC_PAGE_SIZE)) == -1) ||
 	    ((physpages = sysconf(_SC_PHYS_PAGES)) == -1)) {
-		/* Did an error occur? */
-		if (errno != 0)
+		/*
+		 * Did an error occur?  OS X may return EINVAL due to not
```
gperciva: You don't want to wrap this in an `#if OSX` or whatever the appropriate macro is?

cperciva: I considered that, but if other OS kernels return EINVAL I'd want to handle it the same way.  The bug here is in the header files (they shouldn't define _SC_ macros which the kernel doesn't support) rather than in the kernel.

PR #98, opened 2017-01-10
build: get version number from a script

gperciva:
A new version string is generated whenever somebody runs autoreconf.  This
will simplify any downstream releases that do not originate from an official
release tarball.  Inspired by Barak A. Pearlmutter's work:
https://github.com/barak/scrypt/commit/91cb727ba0d66c57d58554258e2de1535a5b71b8


gperciva:
heh, that was fun -- Travis-ci only checks out the last 50 commits, so `git describe` was failing.  I'd made that an error case (since I figured that the developer should investigate), but the error message wasn't visible since I was dumping it to stdout instead of stderr.

get-version.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh
```
gperciva: Not certain if you want this at top-level, or put inside release-tools/.  If the latter, we'll need to distribute that directory.

release-tools/mktarball.sh
```diff
@@ -10,7 +10,7 @@ RELEASEDATE=`date "+%B %d, %Y"`
 
 # Copy bits in
 mkdir ${DESTDIR} ${DESTDIR}/autotools
-cp scrypt_platform.h main.c FORMAT COPYRIGHT ${DESTDIR}
+cp scrypt_platform.h main.c FORMAT COPYRIGHT get-version.sh ${DESTDIR}
```
cperciva: Why?  If we're building a release tarball, it won't have a `.git` directory to extract a version from...

gperciva: Without `get-version.sh`, autotools bails because it's called by `configure.ac`.  But you're right that I forgot that we don't distribute autotools.

I've updated the PR to delete it from `DESTDIR` along with other autotools stuff.  If you like it, I'll merge it with the original commit and rewrite history.

gperciva:
Revised version based on offline discussion.  @barak, do you think this will work for your Debian packages as well?

barak:
Unfortunately
````
if [ -d .git/ ]; then
````
is not a good way to check if you're inside a git repo. Git submodules now sometimes use a regular file for .git containing text specifying the path to the actual repo support directory.

The "approved" thing is to run
````
git rev-parse
````
which will return with status zero if in a git repo (even if deeply nested) and with status non-zero otherwise.

**Comment by barak** [comment]
And if you want to check if you're in the top-level directory of a git repo:
````
git rev-parse --show-cdup
````

**Comment by gperciva** [comment]
Thanks!  I've changed this to use

    if git rev-parse 2>/dev/null; then

since it writes useless text to stderr if it's outside of a git repo.

**Comment by barak** [comment]
Sure, builds fine. I might have to patch the version number stuff to make it build identically inside or outside git (which is a security measure Debian is trying to satisfy for all packages---"reproducible builds") by forcing the version, maybe I'll just hot-wire `get-version.sh`. And the installed man page source contains `@DATE@` which `man` renders as the current  date. (This is a feature I was previously unaware of!) I might add a snippet to switch it to the date in the Debian changelog, `dpkg-parsechangelog --count 1 --show-field date`.

But anyway, sure I can manage.

**Comment by cperciva** [comment]
> make it build identically inside or outside git (which is a security measure Debian is trying to satisfy for all packages---"reproducible builds") 

Wait, what?  I understand wanting reproducible builds -- it's something I was making a fuss over back in 2003 -- but why should it match the git tree?  The git tree is not a release!  The tarball we ship is the release.

**Comment by barak** [comment]
(actually `man` seems to ignore `.Dd` as far as I can tell, uses the current date in the footer regardless)

**Comment by barak** [comment]
Well, we often want to merge in post-release upstream commits etc. Generally the files in the tarball are a superset of what's in git---or at least almost a superset, with maybe some irrelevant things left out. In the present case things are a bit unusual, in that there are a few differences: expanding @DATE@ and moving the autotools stuff into a subdir. Still, it seems desirable to have the generated binary be identical when run out of the tarball and out of git at the corresponding commit. Just because it makes for one fewer thing to check in the event that some problem crops up. I'm not saying it's a big deal. Just a why not make life easy sort of thing.

Also, some build systems in some modes of operation put everything in a new git repo, unrelated to your upstream git repo, prior to building. So I think I should disable "check if you're inside a git repo and build differently if so" to avoid issues in that sort of configuration.

**Comment by gperciva** [comment]
The "check that we have a git directory" is only there to make sure that `git describe` can run, and that the tag matches the version (i.e. `1.2.0-head`).  I suppose I could squash it all onto one line, something like (untested):
```
(git describe --tags --match '[[:digit:]].*' 2> /dev/null && grep -q $( echo ${version} | sed "s/-head//" ) ) || echo ${version}
```
but thought that a script with `if` statements, variables, and multiple lines is much more readable than the above line.  `autoreconf` is not run very often, so there's no speed considerations.

**Comment by barak** [comment]
Well agreed, there is certainly no speed consideration!

But autoreconf is run as part of the default build process in Debian and Ubuntu, and I'd imagine other distributions are moving to do the same. To ease security auditing, among other reasons.

**Comment by gperciva** [comment]
For Debian reproducibility, why not generate your own tarballs?  I mean,
https://packages.debian.org/unstable/main/scrypt
lists
http://http.debian.net/debian/pool/main/s/scrypt/scrypt_1.2.0+git.3.c1a9826.orig.tar.gz
which appears to be a tarball of the git repo without the `.git/` dir.  You could run
```
sh release-tools/mktarball.sh 1.2.0+git.3.c1a9826
```
to generate that tarball instead?

Granted, that tarball wouldn't be cryptographically signed with cperciva's key, but neither is `scrypt_1.2.0+git.3.c1a9826.orig.tar.gz`.  My vague sense from looking at other debian packages is that they normally ship the upstream tarball, sometimes with some patches in `debian/patches` for vital problems.  I wonder if we're inventing problems by trying to have "upstream non-released tarballs".

barak:
The issue is that sometimes the tarball, regardless of how it's generated, will be unpacked into a git repo as part of the build process.

Maybe some environment variable which the version-figuring-out script uses if set? This would allow the build scripts to set the variable to the version string they want, overriding the git repo groveling etc.

gperciva:
Ah, I think I understand now.  Since the tarball *doesn't* contain a `.git/`, any attempt at `git describe` will pick up info from an unrelated repo.

I like the idea of an environment variable; I'll revise the patch accordingly.  I kind-of wish we could ignore autoconf; this kind of problem is exactly why cperciva ships a pre-generated `configure`.

barak:
The autotools system is like democracy, I suppose.

gperciva:
Updated PR to obey to `$SCRYPT_VERSION`.

gperciva:
(going for a "painfully funny" reaction to https://github.com/Tarsnap/scrypt/pull/98#issuecomment-275464663 )

barak:
Good stuff.

I think you're supposed to use `printf '%s' "$..."` in case there's a % in the string to be printed.

````
AX_SANE_PRESIDENT     # take measures to ensure POTUS is sane
````

PR #101, opened 2017-02-10
doc: replace LDADD -> LDFLAGS in OSX instructions

gperciva:


gperciva:
Sorry, I forgot to change one instance of `LDADD` when I copied these from the spiped `BUILDING` file.

BUILDING
```diff
@@ -18,11 +18,14 @@ Platform-specific notes
   10.11 "El Capitan" and higher, OpenSSL was removed entirely.  After
   installing a newer version of OpenSSL, use
       CPPFLAGS="-I /path/to/openssl/headers"
-      LDADD="-L /path/to/openssl/lib"
+      LDFLAGS="-L /path/to/openssl/lib"
   to build spiped.
```
cperciva: It occurs to me that we should probably have s/spiped/scrypt/ here too...

gperciva: whoops.  I also discovered another "spiped" earlier in that paragraph.

gperciva:
PR updated.  I think it's decent now, but I wrecked myself at the gym over lunch so I wouldn't trust my ability to proof-read anything over a grade 2 reading level at the moment.  :)

BUILDING
```diff
@@ -14,15 +14,18 @@ Platform-specific notes
...
-  to build spiped.
+      LDFLAGS="-L /path/to/openssl/lib"
+  to build scrypt.
 
   In particular, if you installed OpenSSL using homebrew, you may
-  add these lines to your $HOME/.profile file:
+  pass the relevant directories directly to ./configure:
+      ./configure CPPFLAGS="-I/usr/local/opt/openssl/include" LDFLAGS="-L/usr/local/opt/openssl/lib"
+
+  Alternately, you may with to add these lines to your $HOME/.profile file:
```
cperciva: Alternatively.  The use of "alternately" to mean "alternatively" is an American perversion.

gperciva: well, I don't think either of those is on the grade 2 reading list.  :)

gperciva:
updated PR

PR #104, opened 2017-02-14
travis-CI: use apt plugin instead of sudo

gperciva:
This looks like a cleaner solution, and gives Travis-CI the
flexibility to use container-based workers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: C
-sudo: required
+sudo: false
 dist: trusty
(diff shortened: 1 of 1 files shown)


gperciva:
Despite https://github.com/travis-ci/apt-package-whitelist/issues/3635, is appears that valgrind is installed.  I checked that it really is installed, and detects memory leaks, with this failing build:
https://travis-ci.org/Tarsnap/scrypt/builds/201395444

(it took a while to find a simple memory leak that clang didn't optimize away)

PR #107, opened 2017-03-29
Parsenum

gperciva:


main.c
```diff
@@ -25,6 +25,7 @@
  */
 #include "scrypt_platform.h"
 
+#include <math.h>
```
gperciva: I added this for INFINITY, used (at the moment) for `-t maxtime`.  I'm not certain if you want an upper bound on that -- sure, if they wanted to not time out, they'd just omit the `-t` parameter, but the original code didn't impose an upper bound.

cperciva:
Supporting INFINITY is good -- not that we particularly want to support INFINITY itself, but INFINITY is the only upper bound we can specify which will be greater than any non-infinite value. :-)

Is the `(double)` cast necessary though?

gperciva:
Yes; that's why `libcperciva/tests/parsenum/main.c` includes it.  Without the `(double)`, `clang` gives us:

```
main.c:111:34: error: implicit conversion increases floating-point precision:
      'float' to 'double' [-Werror,-Wdouble-promotion]
        TEST4_SUCCESS("inf", double, 0, INFINITY, INFINITY);
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
```

`gcc` doesn't complain.  I suppose we could add `-Wno-double-promotion` to `parsenum.h` instead of doing the cast in the `.c` file, though.

cperciva:
Oh, I assumed it was necessary in `TEST4_SUCCESS` due to the test-framework bits, not the parsenum call.  Hmm... is there any reason to not have the explicit `(double)` cast in parsenum itself?

gperciva:
Sketch of a solution in https://github.com/Tarsnap/libcperciva/pull/61.  I'm not wild about one part of it.

gperciva:
Updated PR

PR #114, opened 2017-10-27
Improve tests

gperciva:


gperciva:
Nothing earth-shattering here; mostly just simplifying the non-shared code in order to make it easier to use in a new project.

tests/shared_test_functions.sh
```diff
@@ -42,6 +42,9 @@ out_valgrind="tests-valgrind"
 valgrind_suppressions="${out_valgrind}/suppressions"
 valgrind_suppressions_log="${out_valgrind}/suppressions.pre"
 
+# Print output about test failures.
+VERBOSE=${VERBOSE:-0}
```
cperciva: Just checking... setting VERBOSE to 0 if not defined sounds like it's *not* printing things?  Maybe this just needs the comment clarified.  (e.g., maybe it means "print output about test failures but not test successes" or something like that.)

gperciva: err, yeah.  `VERBOSE=0` means that it only prints `PASS` or `FAIL`.  I'll update the comment.

gperciva:
Updated PR; history is ok for merging if there's no other problems.

Issue #115, opened 2017-11-18
Official website GPG keys for Scrypt


miaarcand:
Hey guys,

What's up with GPG keys for Scrypt on your official website here: https://www.tarsnap.com/scrypt.html ?

```
gpg2 --verify scrypt-sigs-1.2.1.asc scrypt-1.2.1.tgz 
gpg: not a detached signature
```

```
gpg2 --verify tarsnap-signing-key-2017.asc scrypt-1.2.1.tgz 
gpg: verify signatures failed: Unexpected error
```

cperciva:
The `.asc` files are cleartext signatures of the SHA256 output.  They're not detached signatures.

miaarcand:
@cperciva,

I used to verify all GPG-signed files with "gpg --verify".
So, how do I verify Scrypt files?

cperciva:
```
$ gpg --decrypt < scrypt-sigs-1.2.1.asc 
SHA256 (scrypt-1.2.1.tgz) = 4621f5e7da2f802e20850436219370092e9fcda93bd598f6d4236cce33f4c577
gpg: Signature made Fri Feb 10 23:40:33 2017 PST
gpg:                using RSA key DF27AF8EAB663109
gpg: Good signature from "Tarsnap source code signing key (Tarsnap Backup Inc.) <cperciva@tarsnap.com>"

$ sha256 scrypt-1.2.1.tgz 
SHA256 (scrypt-1.2.1.tgz) = 4621f5e7da2f802e20850436219370092e9fcda93bd598f6d4236cce33f4c577
```

miaarcand:
Thanks.

That's the first time I see this. Maybe you want to add a note about verification process to your website?

PR #116, opened 2017-11-18
README.md: clarify how to use scrypt-sigs-*.asc

gperciva:


gperciva:
Note that `gpg --verify` can read directly from a file; no need to pipe stuff.  (response to #115)

Issue #120, opened 2018-02-23
failed decryption creates empty output file


barak:
This is with version 1.2.1. A failed decryption creates an empty output file.
```
$ scrypt dec /dev/null foo-dec
Please enter passphrase: 
scrypt: Input is not valid scrypt-encrypted block
$ ls -l foo*
-rw-rw-r-- 1 barak barak 0 Feb 23 17:12 foo-dec
```
This also happens when decrypting a valid file with the wrong passphrase.

gperciva:
Hi @barak, I've addressed this in #121.

barak:
Cool, thanks @gperciva.

(Unless you instruct me otherwise, I'll wait until you make a stable release before putting this in the Debian package.)

gperciva:
Fixed in #144.

PR #121, opened 2018-02-23
Remove empty file

gperciva:


gperciva:
Addresses #120

cperciva:
What happens if I run `scrypt dec foo /dev/stdout`?  Maybe we should make sure the path we're going to delete is actually a file?

gperciva:
Do we care if somebody plays games with symlinks to try to force us to create a 0-byte file?  i.e.
```
ln -s target link
scrypt dec encfile link
```
then they deliberately enter the wrong passphrase.

I've got code that follows the link and deletes the resulting 0-byte file, but it feels like it's overly complicated and messy.

gperciva:
Updated PR, including questionable commit 9039a74 to handle symlink'd output filenames.

cperciva:
The more I think about this, the more I think deleting files if we fail is a bad idea.  How about checking if the passphrase works before we create the file?  Of course that may result in users getting annoyed if they have to enter the passphrase only to end up being told that the output file can't be created...

PR #123, opened 2018-04-01
Add support for Intel SHA256 instructions

noloader:
This is a partial cut-in of SHA256 support on Intel x86 platforms. I did not see how to glue it into `sha256.c` so others will need to finish the integration (if there is interest in the code).

This implementation will run about 3.8 cycles per byte.

The SHA256 implementation came from [SHA-Intrinsics](https://github.com/noloader/SHA-Intrinsics). All the code is public domain. If scrypt is interested in the code, then similar code is available for ARM and POWER8.

I'm not sure what is going on with the Clang failure on the integral constant. I've been using the code for about a years and a half under Clang, GCC, ICC and MSVC, and it has never caused a problem. A question is open at [Problems with hexadecimal constant, _mm_set_epi64x and sign conversion](https://lists.llvm.org/pipermail/cfe-users/2018-March/001284.html).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -82,6 +82,7 @@ EXTRA_DIST = \
 	lib/README						\
 	README.md						\
 	lib/crypto/crypto_scrypt-ref.c				\
+	libcperciva/cpusupport/Build/cpusupport-X86-SHANI.c	\
 	libcperciva/cpusupport/Build/cpusupport-X86-AESNI.c	\
 	libcperciva/cpusupport/Build/cpusupport-X86-CPUID.c	\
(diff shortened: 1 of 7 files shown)


cperciva:
Thanks for your interest in this!  I prefer to have this sort of thing go into the libcperciva tree first, and I've had an issue open there for a while (https://github.com/Tarsnap/libcperciva/issues/44) -- would you mind taking this and working on it in that tree?

noloader:
Thanks @cperciva,

Yeah, sure.

I'm getting frustrated with Git and its inability to perform simple tasks. I'm going to close this PR and drop a diff at the bug report. Luckily Git can produce diff's easy enough.

Sorry about the extra noise on your bug tracker.

PR #124, opened 2018-04-08
Check passphrase

gperciva:


gperciva:
Quick sketch of checking the passphrase before opening the output file (replaces #121).  Is this approximately what you had in mind?

(I know there's some fine details I need to clean up)

cperciva:
This results in the key derivation being done twice.  How about splitting `scryptdec_file` into `scryptdec_file_open(infile, passwd, passwdlen, maxmem, maxmemfrac, maxtime, verbose, force)` which returns a cookie (containing the derived key) or NULL; and `scryptdec_file_copy(infile, cookie, outfile)` which performs the decryption?

PR #125, opened 2018-04-11
Check passphrase cookie (prep)

gperciva:

PR #126, opened 2018-04-12
Get passphrase before opening output file

gperciva:


gperciva:
*sigh* I wish that github wouldn't get confused if I force-push a branch before remembering to re-open the related PR.

Anyway, this is some prep for the `scryptdec_file_open()` and `scryptdec_file_copy()` change suggested in #125.  I'm taking it very cautiously because every time I made "one last paranoid check" today, I discovered a new way that I was failing to zero the passphrase.

I'm reasonably confident that this small first step is correct, but I'd appreciate you checking it before I upload more of my work.

cperciva:
Please put imports from libcperciva into their own PR separate from the non-libcperciva changes.

main.c
```diff
@@ -233,6 +233,14 @@ main(int argc, char *argv[])
 	/* Success! */
 	return (0);
 
+err1:
+	/* Zero and free the password. */
```
cperciva: I don't normally repeat comments like this in the error path -- it doesn't matter so much here, but it's often annoyingly repetitive when exactly the same comments appear in the "success" cleanup path too.

gperciva:
Updated PR

PR #127, opened 2018-04-23
Libcperciva import

gperciva:


gperciva:
Is it helpful to know the hashes from these commits in libcperciva?  I could add them if you want, but if it's not important then of course I won't bother.

cperciva:
Nah, people can find changes in the libcperciva tree easily enough without needing hashes.

PR #128, opened 2018-04-23
tests: add "decrypt with bad passphrase"

gperciva:


gperciva:
Split off from #126 since it's not as connected as I thought at first.

(I'd previously mis-thought that `readpass()` also checked the validity of the passphrase, which is of course nonsense.  So this test doesn't depend on having an error exit path.)

PR #132, opened 2018-05-08
Sanitize derived key from stack on error

gperciva:


cperciva:
Hmm, looks like it would be cleaner if `scryptdec_setup` doesn't zero the buffer but instead its caller handles the zeroing?  I mean, the buffer `dk` is a parameter to `scryptdec_setup`; it's not a buffer which that function allocated itself.

gperciva:
PR updated with `scryptdec_setup()` not zeroing.  I added a comment warning about that function potentially having a tainted `${dk}` even if it quits with an error, since I'd find that surprising.  I expect that you'd prefer a different warning to what I added.

cperciva:
I would go with `/* NOTE: The caller is responsible for sanitizing ${dk}, including if this function fails. */` and also add a similar comment to `scryptenc_setup`.

Also, now that `scryptdec_setup` isn't zeroing `dk`, you'll need to do that in the code which calls that function if it fails.

gperciva:
... huh, I was going to ask if `crypto_scrypt()` can really leave sensitive data in memory (because otherwise, `scryptenc_setup()` wouldn't need any sanitizing upon failure), but I see that `munmap()` can fail at the end of `_crypto_scrypt()`.

`scryptdec_setup()` already has `goto err1;` to sanitize `dk`, changed in the first `REBASE` commit.

cperciva:
Looks good; please clean up the commit history.

gperciva:
Updated PR, ready for merge.

PR #133, opened 2018-05-14
Libcperciva import

gperciva:


gperciva:
This also includes the whitespace fixes in https://github.com/Tarsnap/libcperciva/pull/119, but I don't anticipate any reason why that wouldn't be merged.

PR #134, opened 2018-05-15
Don't write an empty output file if given a bad decryption phassphrase

gperciva:


gperciva:
This is a large PR, which I've broken down into 3 separate PRs.
- #132 add error path to `scryptdec`, which will be merged soon.
- #135 add a `scryptdec_file_cookie`.
- #136 refactor `int main()`.

Don't bother looking for nitpicks in this one yet; I'm putting it here so you can see how the other PRs fit together, and so you can decide if the overall design is ok.

gperciva:
This is my preferred solution (instead of #141); I think that the code is much cleaner and easier to read.

If you agree, I suggest evaluating #136 by itself first.

PR #135, opened 2018-05-15
Scryptdec cookie internal

gperciva:


gperciva:
This PR includes #132 (which will probably be merged soon), and is part of #135.

lib/scryptenc/scryptenc.c
```diff
@@ -664,32 +719,63 @@ scryptdec_file(FILE * infile, FILE * outfile,
...
+int
+scryptdec_file(FILE * infile, FILE * outfile, const uint8_t * passwd,
+    size_t passwdlen, size_t maxmem, double maxmemfrac, double maxtime,
+    int verbose, int force)
+{
+	struct scryptdec_file_cookie * C;
+	int rc;
+
+	/* Check header, including passphrase. */
+	if ((rc = scryptdec_file_prep(infile, passwd, passwdlen, maxmem,
```
cperciva: I would write this as `if ((rc = ...) != 0)` since I think that's clearer (especially to people who have trouble counting parentheses).

gperciva:
Updated; shall I rebase?

cperciva:
I feel like there should be more clarity about ${infile}; in particular, that it is not freed by these functions upon error, but that the caller must not change the file position in between them (since the file is kept positioned after the end of the header).

Aside from that, please rebase.

gperciva:
rebased but more REBASE commits added.

- instead of adding warning that the caller should not change the position of `${infile}`, I added a `ftell(); fseek()` combo.  That seemed like a better (safer) API design than a mere warning comment?

- since `${infile}` isn't allocated inside this `.c` file, I'm not certain why anybody would think that these functions would free it upon an error?  Anyway, I added a warning to the function comments.  If you're sure you want that approach, I'll do it properly (i.e. add it to the `.h` file for `scryptdec_file()` as well, make sure each function gets that comment at the appropriate stage of the commit history, etc.).

gperciva:
Updated with REBASE commits about clarifying ownership of `${infile}`.

cperciva:
Looks good, please rebase.

gperciva:
Updated PR with rebased history, ready for merge.

gperciva:
I just noticed that we probably want a

    assert(C != NULL)

in `scryptdec_file_prep()`.

gperciva:
... and by `_prep`, I mean `_copy`.  Added and rebased into history.

lib/scryptenc/scryptenc.c
```diff
@@ -674,10 +671,8 @@ scryptdec_file(FILE * infile, FILE * outfile, const uint8_t * passwd,
 	struct crypto_aesctr * AES;
 	int rc;
 
-	/* Check header, including passphrase. */
-	if ((rc = scryptdec_file_prep(infile, passwd, passwdlen, maxmem,
-	    maxmemfrac, maxtime, verbose, force, &C)) != 0)
-		goto err0;
+	/* Sanity check. */
+	assert(C != NULL);
```
gperciva: These two lines are the added sanity check in `scryptdec_file_copy()`.

PR #136, opened 2018-05-15
Main refactor

gperciva:


gperciva:
This is part of #134.  Right now having `open_output()`, `encrypt_file()`, and `decrypt_file()` looks rather over-engineered, but it makes it much cleaner to start using `scryptdec_file_cookie` in https://github.com/Tarsnap/scrypt/pull/134/commits/9d0bc1669a05e81da9ca3665bc9d645a6d0f4a4e.

main.c
```diff
@@ -224,7 +231,7 @@ main(int argc, char *argv[])
 			    (argc > 1) ? argv[1] : "standard output");
 			break;
 		case 13:
-			warnp("Error reading file: %s", argv[0]);
+			warnp("Error reading file: %s", infilename);
```
cperciva: What happens if we have trouble reading stdin?

main.c
```diff
@@ -192,28 +261,22 @@ main(int argc, char *argv[])
...
 		goto err1;
 
-	/* If we have an output file, open it; otherwise, use stdout. */
-	if ((outfile = open_output(outfilename)) == NULL)
-		goto err2;
-
 	/* Encrypt or decrypt. */
 	if (dec)
-		rc = scryptdec_file(infile, outfile, (uint8_t *)passwd,
+		rc = decrypt_file(infile, outfilename, (uint8_t *)passwd,
```
cperciva: Not sure what this change accomplishes aside from duplicating lots of code?

gperciva:
This refactoring allows us to add #134 (using `scryptdec_file_prep()` and `scryptdec_file_copy()`) in a much cleaner fashion.  It's the first step towards that cleaner implementation, instead of the mess that is #141.

PR #140, opened 2018-08-29
Main with filenames

gperciva:


gperciva:
I get confused when looking at stuff like
```
warnp("Error writing file: %s",
               (argc > 1) ? argv[1] : "standard output");
```

This is split off from #136.  (I'm working on an alternate PR to that which doesn't add extra functions, but I'd still like to include these two commits.  It'll be easier to compare #136 and the new PR if these two commits are already in master.)

main.c
```diff
@@ -224,7 +231,7 @@ main(int argc, char *argv[])
 			    (argc > 1) ? argv[1] : "standard output");
 			break;
 		case 13:
-			warnp("Error reading file: %s", argv[0]);
+			warnp("Error reading file: %s", infilename);
```
cperciva: What if we have trouble reading stdin?  (Entirely possible if someone runs `scrypt enc - < /file/on/disk` and there's an I/O error.)

gperciva: Rebased and pushed a new version that uses the "Error writing file" approach of checking the variable and printing "standard input" if appropriate.

PR #141, opened 2018-08-30
Scryptdec cookie (alternate solution: all in main)

gperciva:


gperciva:
This is an alternate PR that would replace #134 and #136.

I don't like it because the error paths through `main()` are confusing.  This version passes the test suite, but I'm not 100% certain that it's doing the right thing in all cases.  I'm particularly unhappy with the error path for `scryptdec_file_prep()` -- setting `C = NULL` isn't necessarily terrible, but `outfile = stdout` is really questionable.

I think that once you compare this to #134, you'll agree that #134 is better.

cperciva:
This seems cleaner to me than #134.  It's certainly a more minimal diff (+60 -12, vs. +135 -22).

If you don't see any problems with this (feel free to tidy up the commit history) I'll merge this and close #134 and #136.

gperciva:
I'm surprised that you're going with `outfile = stdout`, because you generally don't like setting default values.  If we're making that change, might I also suggest:
- `infile = NULL`
- `infilename = outfilename = NULL`

as well?

cperciva:
> How about downwithfiles?

Down with files!  Power to the sockets!

gperciva:
Aren't sockets pretending to be files?  So down with them, too!

... I mean, we *are* supposed to call `shutdown()` when we're done, right?

PR #142, opened 2018-08-31
Libcperciva import

gperciva:


gperciva:
I tried to check the SHANI code by running it on a SHANI system, but the system we formerly used is unavailable (`gcc68` "unreachable for 28 days": https://cfarm.tetaneutral.net/machines/list/).  I'm not concerned that the code magically stopped working -- we tested it at the time -- it's just that I imagined that you'd ask if I'd double-checked that it worked in scrypt.

I could split up the SSSE3 and SHANI imports into separate commits if you want.

PR #143, opened 2018-09-14
Scryptdec cookie all in main rebased

gperciva:

PR #144, opened 2018-09-14
tests: "dec with bad passphrase" shouldn't create an output file

gperciva:


gperciva:
Resolves #120 (well, #143 actually resolved it, but this test should make sure that the solution doesn't break in the future).

PR #147, opened 2018-09-16
Fix library name

gperciva:


gperciva:
I don't know how committed we are to the odd sorting order in `scrypt_SOURCES`.  If we do anything non-obvious (e.g., "start with `main.c`, then purely alphabetical") then we should document what it is that we want.

cperciva:
I think my intention was "main.c, then libcperciva in alphabetical order, then lib in alphabetical order".  That sort of went out the window when you added all of the `.h` files to `scrypt_SOURCES` though.

But I'm fine with any ordering really.  Should I merge this as-is, or do you want to rearrange things?

gperciva:
Rearranged / rebased to have `main.c` first, and then alphabetical.  I like alphabetical, because I can do `:'<,'>sort` and that takes thinking out of that part of build-file maintenance.  I like not thinking.

PR #148, opened 2018-09-23
Build: shared files for crypto_scrypt

gperciva:


gperciva:
I think this is useful a clean-up for the build system, regardless of whether we end up doing `libscrypt` or not.  (I'm still working on making `libscrypt` clean enough to propose)

Makefile.am
```diff
@@ -2,27 +2,38 @@ bin_PROGRAMS=	scrypt
 noinst_PROGRAMS=	tests/test_scrypt tests/valgrind/potential-memleaks
 dist_man_MANS=$(scrypt_man_MANS)
 
+# These files fulfill the crypto_scrypt() function, with the exception of
```
gperciva: I spent a minute agonizing over whether it should be "fulfill", or "provide", or "implements", or something else.  I think the meaning is clear in context (particularly if you look at `test_scrypt`), but I also think that "fulfill" isn't quite the right word.

gperciva:
Wait, I suspect that we should include `cpusupport.h` in there.  I'm still trying to work my mind around headers in `_SOURCES` in automake.

gperciva:
Updated to include `cpusupport.h` and `scrypt_platform.h`.

PR #153, opened 2018-12-25
cpusupport: import CPUID_COUNT

gperciva:


gperciva:
Without this, `CPUSUPPORT_X86_CPUID_COUNT` will not be set, and as a result, `CPUSUPPORT_X86_SHANI` won't work.

Very few CPUs have SHANI instructions, which is why we didn't notice this earlier.

PR #154, opened 2018-12-28
MSVC compatibility

ml1nk:
As asked in https://github.com/ml1nk/node-scrypt/issues/8 this pull integrates the changes made in ml1nk/node-scrypt for scrypt-1.2.1 to compile it with MSVC.

In scrypt 1.2.0 no changes were necessary but since 1.2.1 it does not compile without these changes due to two incompatibilities with MSVC:
- sys/time.h (gettimeofday) does not exist
- static restrict is not supported

Reproduction:
Install Node.js with build tools on Windows and clone https://github.com/ml1nk/node-scrypt.
Replace scrypt/scrypt-1.2.1 with a current version with and without the fixes.
Try to compile the code with MSVC by executing npm install in the base folder of node-scrypt.

### DIFF
--- a/lib/scryptenc/scryptenc_cpuperf.c
+++ b/lib/scryptenc/scryptenc_cpuperf.c
@@ -28,7 +28,12 @@
  */
 #include "scrypt_platform.h"
 
+#ifndef _MSC_VER
 #include <sys/time.h>
+#else /* For 'struct timeval' and custom gettimeofday() on windows */
+#include <winsock2.h>
+#include "gettimeofday.h"
+#endif /* _MSC_VER */
 
 #include <stdint.h>
 #include <stdio.h>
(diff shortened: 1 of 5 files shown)


gperciva:
Thanks for sending these patches upstream!

It might be tricky to accept these, though.  Our software is written for the C99 and POSIX standards.  We can work around some POSIX incompatibility, but if the compiler can't handle C99 then things can get really icky.  Do you have gcc or clang available?

I'm currently looking at alternate ways of dealing with `scryptenc_cpuperf.c`; I'm cautiously optimistic that we can do something here.  I'm less optimistic about avoiding `static restrict`, though... but of course it's @cperciva that has the final say, not me.

cperciva:
Yeah, I'm inclined to say that MSVC is just too far off to be something we're going to make an effort to support.  Thanks for understanding!

PR #155, opened 2019-01-01
Document our use of POSIX

gperciva:


gperciva:
Imports the text we use in spiped `BUILDING`.  Arose from #154, because I realized that we didn't specify that we use POSIX anywhere in the scrypt repository.

gperciva:
... of course, that assumes that you deliberately want to depend on POSIX.  It's just possible that you wanted scrypt to be (theoretically) usable on non-POSIX systems.

PR #156, opened 2019-01-03
Use monoclock for scryptenc

gperciva:


Makefile.am
```diff
@@ -64,7 +66,9 @@ AM_CPPFLAGS=		-I$(srcdir)/libcperciva/alg		\
 			-I$(srcdir)/lib/scryptenc 		\
 			-I$(srcdir)/lib/util			\
 			-DCPUSUPPORT_CONFIG_FILE=\"cpusupport-config.h\" \
-			-D_POSIX_C_SOURCE=200809L
+			-D_POSIX_C_SOURCE=200809L		\
+			-D_XOPEN_SOURCE=700			\
```
gperciva: Technically we don't need `XOPEN_SOURCE` in here, but our other software projects use it, so I added it in the interest of simplicity.

lib/util/memlimit.c
```diff
@@ -29,6 +29,7 @@
 
 /* We use non-POSIX functionality in this file. */
 #undef _POSIX_C_SOURCE
+#undef _XOPEN_SOURCE
```
gperciva: The version of this file in the tarsnap client code already has this `undef`.

gperciva:
You might want me to merge the last two commits.

This PR is blocked on a trivial bugfix -- `monoclock.c` currently has an unused `err0:` if we're not using `CLOCK_MONOTONIC` or `CLOCK_REALTIME`.

cperciva:
Yes, you can merge "use new code" with "get rid of old code".

BTW can we stop asking autoconf to look for clock_gettime now?

gperciva:
Updated PR.

Just like the tarsnap client, we're only using the CFLAGS compatibility layer, so we still rely on autoconf to figure out if it needs `-lrt` or not.  I suppose that we could remove the `AC_CHECK_FUNCS([clock_gettime])`; should I investigate that?

cperciva:
Oh right, I was forgetting that the CFLAGS and LDFLAGS bits were separate.

This is not at all urgent, but my general preference is to rely on autoconf for as little as possible, so if you feel like mucking about with autoconf feel free to investigate.

cperciva:
Is this ready to merge?  It's still marked as "draft".

gperciva:
Oops, yes, good to merge now.

gperciva:
No wait, let's get LDADD in at the same time.  I think we should merge https://github.com/Tarsnap/scrypt/pull/167 first.

gperciva:
Rebased onto master, and finally ready for merge AFAIK.

PR #157, opened 2019-01-03
Move tests/test_scrypt.c into tests/verify-strings/

gperciva:


gperciva:
The `tests/` directory was getting a bit messy, so I moved `tests/test_scrypt*` into a separate directory.

Issue #131, opened 2018-04-26
Available memory on FreeBSD


msimonsson:
Hi,

I had a little scare yesterday trying to decrypt backups on my home server. Every attempt resulted in a "Decrypting file would require too much memory".

Here is the output from top:

> last pid: 53207;  load averages:  0.25,  0.18,  0.13  up 20+23:19:18  14:35:06
> 54 processes:  1 running, 53 sleeping
> CPU:  0.0% user,  0.0% nice,  0.0% system,  0.0% interrupt,  100% idle
> Mem: 19M Active, 258M Inact, 23M Laundry, 14G Wired, 774M Free
> ARC: 13G Total, 11G MFU, 664M MRU, 32K Anon, 70M Header, 477M Other
>      12G Compressed, 13G Uncompressed, 1.14:1 Ratio
> Swap: 4096M Total, 85M Used, 4011M Free, 2% Inuse

ZFS ARC is using almost all available memory, but most of this memory should be available to other processes when needed.

I found the "force" option for decryption but if I encrypt a file under these conditions wouldn't scrypt pick less than ideal memory parameters?

Thanks,
Mikael

cperciva:
Hmm, interesting.  Can you compile scrypt with DEBUG defined (just add `#define DEBUG` to the top of `lib/util/memlimit.c`) and tell me what it prints?  I'm curious what values it is coming up with in this case.

msimonsson:
Sure, here's a before and after:

> last pid: 85679;  load averages:  0.11,  0.13,  0.09
> 46 processes:  1 running, 45 sleeping
> CPU:  0.0% user,  0.0% nice,  0.0% system,  0.0% interrupt, 99.9% idle
> Mem: 12M Active, 235M Inact, 8871M Wired, 6735M Free
> ARC: 2607M Total, 1077M MFU, 1028M MRU, 55K Anon, 34M Header, 468M Other
>      1745M Compressed, 3924M Uncompressed, 2.25:1 Ratio
> Swap: 4096M Total, 4096M Free
> 
> Memory limits are 7801532416 18446744073709551615 18446744073709551615 34359738368 17103745024
> Allowing up to 975191552 memory to be used

> last pid: 86331;  load averages:  0.99,  0.52,  0.27
> 50 processes:  2 running, 48 sleeping
> CPU:  6.6% user,  0.0% nice,  2.4% system,  0.0% interrupt, 91.0% idle
[...]

PR #164, opened 2019-04-07
memlimit: avoid bogus limit arising from ZFS

gperciva:


gperciva:
Addresses #131.

I think that you want this after the `#ifdef DEBUG` that prints memory limits, but it's easily moved if you'd prefer it to be before.

PR #167, opened 2019-04-18
Import posix

gperciva:


gperciva:
Needs more ~cowbell~ autotools `EXTRA_DIST`.

gperciva:
I think this is plausible now.  Tested it (from a tarball) on multiple OSes.

PR #168, opened 2019-04-18
libscrypt-kdf

gperciva:


gperciva:
I couldn't find a nice way to have optional shared libraries.  Autotools really wants you to use libtool, but I couldn't hide `LT_INIT` inside an `AS_IF()` to make it an optional thing.  (And I know that @cperciva wouldn't want to require libtool.)  I considered having an entirely separate `configure.ac` and `Makefile.am` is a separate directory, but I ran into difficulties making that play nice with out-of-tree builds.

I decided that producing a static library was fine for a first step.  If there's a desire to add a shared library at a later stage, hopefully an autotools wizard will step forward.

gperciva:
From offline communication, libtool is ok.  I've therefore reworked this to use libtool, in order to provide a shared library as well.

Two remaining questions:
- should we provide a way to query CPU-specific code?  (something like `cpusupport_x86_shani()`, although reworked to not require `CPUSUPPORT_` defines)
  That could plausibly influence the choice of N, r, p.
- should we provide a convenience wrapper function, such as `insecure_scrypt()` (which has hard-coded values for N, r, p, and possibly takes care of the salt as well)?
  I expect that the answer to this one is **no**, but it's something I saw in downstream projects, so I figured it would be good to have a pre-emptive official ruling.

gperciva:
Note: before the first release that includes libscrypt, we should update `-version-info 0` in `Makefile.am`.

cperciva:
Dumb question: Is it possible to make libtool necessary only for people who want libscrypt and not for people who only want the utility?

gperciva:
I'll check again, but I tried hiding LT_INIT inside an AS_IF but it still generated the configure checks.  It's just possible that I forgot to `:w` the file before running `autoreconf -i` (hence the double-check), but my general impression is that autotools really, really, really wants you to use libtool, so I wasn't surprised when it didn't work.

gperciva:
Looks like I wasn't wrong.  If `configure.ac` contains:
```
AM_COND_IF([LIBSCRYPT],
        [LT_INIT], [AC_PROG_RANLIB])
```

then only one of those **should** be executed.  But instead, I get:

```
$ autoreconf -i
libtoolize: Remember to add 'LT_INIT' to configure.ac.
libtoolize: 'AC_PROG_RANLIB' is rendered obsolete by 'LT_INIT'
```

which amusingly seems to imply that `LT_INIT` is a cat in a box.

Looking at the resulting `configure`, it looks like it contains both LT_INIT and RANLIB stuff.

gperciva:
Rebased onto master

cperciva:
Sigh... the wonders of autoconf...

Ok, we definitely need to make this version 1.3.0.

gperciva:
Not that I disagree with the number 1.3.0, but it just occurred to me that I only tested "does it check for libtool?", not "does it bail if it doesn't find libtool?".  Also, I was looking at the "build from git", not "build from a release tarball".

I'll test the latter now.

cperciva:
Hmm, I wonder if it makes sense for this to be `crypto_scrypt.h` and `crypto_scrypt()`.  Within our tree that namespacing makes sense, but for someone else who is just linking with `-lscrypt` maybe it would make more sense for this to be `scrypt.h` and `scrypt()`?

gperciva:
First thought: if somebody doesn't check the header file, they're not going to know which parameters are for the salt, N, r, etc.

That said, it's quite plausible that somebody would look for a `scrypt.h` file.  But I'd rather not rename the file upon install, since that might cause confusion if somebody looks for it in our git repository.

What would you feel about a symlink for `scrypt.h` ?

cperciva:
I was thinking maybe that we could simply have a separate `scrypt.h` which gets used for libscrypt.  Keeping it in sync with `crypto_scrypt.h` doesn't seem like it would be too hard given how often that file changes.

gperciva:
... ok.  I'm guessing that file would be
```
#ifndef SCRYPT_H
#define SCRYPT_H
#include "crypto_scrypt.h"
#endif
```

cperciva:
No, because I'd prefer to install just one header file, not two. ;-)

And it needs some magic so that people can call `scrypt()` and end up running `crypto_scrypt` code.

gperciva:
ok, I'll add that.

Thinking about release plans... how about we have 1.2.2 without libscrypt, but add a note saying that we're adding a libscrypt library and expect to have 1.3.0 in a week or two.  (or maybe that should be a separate message)

Basically, get the current code out there (in case libtool turns out to be a huge pain for some people), and attract more eyeballs on the libscrypt branch before we have a "libscrypt 1.0.0" out there.

gperciva:
Good news: `libtool` is not at all a requirement for compiling an official tarball.  I tried it on Ubuntu and freebsd; the latter only had:

```
$ pkg info
pkg-1.10.5_5                   Package manager
vim-console-8.1.0985           Improved version of the vi editor (console only)
$
```

so there's absolutely no dependency on autotools.

libscrypt/scrypt.c
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Return 0 on success; or -1 on error.
+ */
+int
+scrypt(const uint8_t * passwd, size_t passwdlen,
+    const uint8_t * salt, size_t saltlen, uint64_t N, uint32_t _r, uint32_t _p,
+    uint8_t * buf, size_t buflen)
+{
+
+	return (crypto_scrypt(passwd, passwdlen, salt, saltlen, N, _r, _p,
```
gperciva: In a comment, you said:
> And it needs some magic so that people can call `scrypt()` and end up running `crypto_scrypt` code.

Do you actually want magic (something fancy), or just a plain function call?

libscrypt/scrypt.h
```diff
@@ -0,0 +1,47 @@
...
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ *
+ * This file was originally written by Colin Percival as part of the Tarsnap
+ * online backup system.
+ */
+#ifndef SCRYPT_H_
```
gperciva: Oops, I took off the initial underscore, but forgot about the trailing one.

BTW: can we not have an initial underscore here?

tests/libscrypt/Makefile
```diff
@@ -0,0 +1,4 @@
+.POSIX:
+
+all:
+	${CC} sample-libscrypt.c -lscrypt
```
gperciva: I forgot that I gave a sample compile command in the README.  We probably don't need this makefile, then.

gperciva:
I'm leaning towards making a `libtool` PR, which switches to that but doesn't introduce any libscrypt stuff.  But I tend to prefer multiple small PRs.

libscrypt/scrypt.c
cperciva: I was thinking that `#define scrypt crypto_scrypt` along with the declaration of `crypto_scrypt` would suffice.

libscrypt/scrypt.h
cperciva: Sure.

gperciva:
Updated PR with rebase

gperciva:
Updated with rename to `scrypt-kdf`.

cperciva:
LGTM, should I merge this now?

gperciva:
I've just rebased it onto master so that we'll have nicer-looking history.  Then it's good for a merge.  :)

gperciva:
actually, I'll do a final test of this on OSX.

gperciva:
Good to go now.  We'll assume that if somebody is installing the library on OSX to a non-standard location, they know how to set up `C_INCLUDE_PATH` etc.

PR #171, opened 2019-05-01
Use libtool for libraries

gperciva:


gperciva:
This is a prerequisite for #168.

PR #172, opened 2019-05-02
build: rename scrypt_platform.h -> platform.h

gperciva:


lib/crypto/crypto_scrypt-ref.c
```diff
@@ -26,7 +26,7 @@
  * This file was originally written by Colin Percival as part of the Tarsnap
  * online backup system.
  */
-#include "scrypt_platform.h"
+#include "platform.h"
```
gperciva: I just noticed that this line can be deleted entirely -- there's no system-specific `#ifdef`s here.  (at least none that use `config.h`)

gperciva:
If this is merged, my plan is to add `platform.h` to the tarsnap client and change the relevant `lib/` files to use that.  But I'll leave `bsdtar-platform.h` alone, because the libarchive stuff obviously expects it.

This does mean that we'll end up with 7 lines duplicated in `platform.h` and `bsdtar-platform.h`, but disk space is cheap these days.  :)   Seriously though, I think that duplication is worth it in order to have a distinction between the nice clean libcperciva/scrypt code, vs. the libarchive code.  I mean, code like `memlimit.c` really doesn't need to include `archive.h` or `dirent.h`!

cperciva:
I agree with not wanting to pollute scrypt code with bsdtar-related headers.

Is it straightforward to have two separate headers which are generated by autoconf?  Or would `platform.h` and `bsdtar_platform.h` both be including the same `config.h` file?

gperciva:
... I was going to say "oh sure, they'll both include the same `config.h`, but that's no problem because of course there's a header guard.

But there isn't.

WTF, autotools?!?!

...

I'll play with this change in the tarsnap client first, and we can import whatever works into scrypt/

gperciva:
Draft of this applied to the tarsnap client: https://github.com/Tarsnap/tarsnap/pull/354

In summary, for the client:
- `.c` files will include `bsdtar_platform.h`, or `platform.h`, or none.  But not all of the above.
- `tar/bsdtar.h` is messy because it includes `bsdtar_platform.h`.  However, none of the files which include `bsdtar.h` would normally include `platform.h`, so we don't need to get into that.

For scrypt,
- this PR is good
- that said, I'd like to merge the libtool changes in #171, then rebase this PR on top of that.

PR #175, opened 2019-05-10
tests/libscrypt-kdf: trivial style fix

gperciva:


gperciva:
I suppose that it was inevitable that I'd forget to run my "find style violations" script.  :(

cperciva:
If we're going to make this stylish, how about giving `main` some arguments?  We'll ignore them, but they should be there...

gperciva:
really?  hmm... is `main(void)` allowed?  We use that a bunch in `libcperciva/POSIX/`.

cperciva:
I think `C99` allows both `int main(void)` and `int main(int argc, char * argv[])` but not `int main()`.  But check the standard to make sure...

gperciva:
Yeah, the spec says `int main(void)` or the usual `argc, argv` thing.  Updated PR.

PR #176, opened 2019-05-20
build: fix dir of sample-libscrypt-kdf

gperciva:


gperciva:
I thought that I'd caught all the `libscrypt` -> `libscrypt-kdf`, but apparently not.

PR #178, opened 2019-06-01
Move to lib/platform.h, plus nitpicks

gperciva:


gperciva:
I like the idea of `lib/platform.h` that you suggested in https://github.com/Tarsnap/tarsnap/pull/354#issuecomment-497920281.  While I was at it, I fixed a few nitpicks.

PR #179, opened 2019-06-04
BUILDING: mention libtool for building from git

gperciva:


gperciva:
For emphasis: `libtool` is only needed when building from git, not the release tarball.

PR #180, opened 2019-06-07
Match function comments between h and c files

gperciva:


libcperciva/crypto/crypto_entropy.c
```diff
@@ -200,7 +200,8 @@ generate(uint8_t * buf, size_t buflen)
 
 /**
  * crypto_entropy_read(buf, buflen):
- * Fill the buffer with unpredictable bits.
+ * Fill the buffer with unpredictable bits.  The value ${buflen} must be
+ * less than 2^16.
```
gperciva: Do we want an `assert` for this?

cperciva: Hmm, I can't remember why I put that restriction there.  Can you check through the history?  It's possible that we can only generate 2^16 bytes before reseeding but I loop around that so we don't actually have this limitation.

gperciva: The discrepancy dates back to spiped 0.9.9.1 on 2011-07-02.  In `lib/crypto/crypto_entropy.h` (this was before you had a `libcperciva/` directory), the `crypto_entropy.c` file contained:

```
/**
 * generate(buf, buflen):
 * Fill the provided buffer with random bits, assuming that reseed_counter
 * is less than RESEED_INTERVAL (the caller is responsible for calling
 * reseed() as needed) and ${buflen} is less than 2^16 (the caller is
 * responsible for splitting up larger requests).  (Section 10.1.2.5)
 */
static void
generate(uint8_t * buf, size_t buflen)
```

Note: that function comment appears to be identical to the current one in `crypto/crypto_entropy.c`.  That file currently contains
```
/* Limited to 2^16 by specification. */
#define GENERATE_MAXLEN 65536
```

The section refers to the NIST SP 800-90 standard (which has been superseded by a few revisions, but I haven't looked at their changelog).  Section 10.1.2.5 Generating Pseudorandom Bits Using HMAC_DRBG says
```
The values for max_number_of_bits_per_request and outlen
are provided in Table 2 of Section 10.1.
```

but table 2 lists `max_number_of_bits_per_request` as `≤ 2^19 bits` (regardless of whether you're doing SHA-1 through SHA-512).  So from casual browsing, it's not clear where the 2^16 comes from, but hopefully this will be enough to jog your memory.

gperciva: I'm an idiot.  2^19 bits is 2^16 bytes.

So yes, the specification introduces a limit.  But no, that limit does not apply to `crypto_entropy()`; it applies to `generate()` instead, which hands the reseeding.  I'm preparing a separate PR for this, with longer reasoning there.

gperciva:
I removed the changes to `elasticarray`, since we'll have a bunch of those coming in an `import libcperciva` PR.

Only a trivial comment-change here; ready for merge.

PR #187, opened 2019-07-01
libscrypt: set version to 1.0.0

gperciva:

PR #188, opened 2019-07-01
Fix libscrypt symbols

gperciva:


gperciva:
This includes #187 because otherwise we'd get merge conflicts.

gperciva:
Wait, that only fixes the `.so`, not the `.a`.

gperciva:
... but it seems that this is normal for static libraries.  So I think this is good to merge.

PR #193, opened 2019-07-10
scrypt info encfile

gperciva:


gperciva:
I'm not certain if this is worth including in the main branch.  I personally would find it useful while working on some scrypt self-tests, and possibly when debugging unexpected scrypt behaviour on some platforms (likely arising from memory or cpu limit detection).  But that could be accomplished with this code remaining on a separate branch.

Sample usage:

```
$ ./scrypt dec --print-params-only config.h.enc 
N, r, p:	262144	8	1
```

(that said, I kind-of like the refactoring commit, so even if we don't add `scryptdec_file_printparams()`, how would you feel about the refactor?)

lib/scryptenc/scryptenc.c
```diff
@@ -68,17 +68,30 @@ display_params(int logN, uint32_t r, uint32_t p, size_t memlimit,
...
-	    "(limit: %.1f seconds).\n", expected_seconds, maxtime);
 
+	/* Memory */
+	fprintf(stderr, "    This requires at least %s bytes of memory",
+	    human_mem_minimum);
+	if (memlimit > 0)
+		fprintf(stderr, " (%s available)", human_memlimit);
+
+	/* CPU time */
+	/* CPU time */
```
cperciva: ?

main.c
```diff
@@ -47,6 +47,7 @@ usage(void)
 	    "usage: scrypt {enc | dec} [-f] [-M maxmem]"
 	    " [-m maxmemfrac]\n"
 	    "              [-t maxtime] [-v] [-P] infile [outfile]\n"
+	    "       scrypt dec --print-params-only infile\n"
```
cperciva: I'd be more inclined to make this `scrypt info infile`; it seems odd to say "decrypt, except don't actually decrypt" when we just mean "print information about this encrypted file".

gperciva: done

gperciva:
Quite a few bits of changed history, and added a man-page entry.

main.c
```diff
@@ -164,6 +167,25 @@ main(int argc, char *argv[])
 		infile = stdin;
 	}
 
+	/* User selected 'info' mode. */
+	if (info) {
+		/* Only valid if we've been given a file. */
+		if (!infilename) {
+			if (infilename != NULL)
```
cperciva: Umm... if `!infilename`, why do we need to check if `infilename != NULL`?

cperciva:
Good aside from the one bit of weird code.  Fix that then I'll merge.

gperciva:
Whoops.  I simplified that code to:
```
                /* Bail if we're looking at stdin. */
                if (infile == stdin)
                        usage();
```

which is also shorter and explains itself much better.

cperciva:
Ok, now that the code makes sense and I know what you're trying to do... why?  Is it impossible that someone might like useless cats and want to do `cat foo | scrypt info -`?  Or perhaps more usefully, "cat foo | ssh host.with.scrypt.installed scrypt info -`?

gperciva:
inb4 "there's no such thing as a useless cat"...

... hmm, I think that I was confused with the "`-P` and `stdin`" thing, or simply with the `fclose()`.  I've removed that check now (and added one to the `fclose()`).

main.c
```diff
@@ -171,6 +174,19 @@ main(int argc, char *argv[])
...
+	if (info) {
+		/* Print the encryption parameters used for the file. */
+		rc = scryptdec_file_printparams(infile);
+
+		/* Clean up. */
+		if (infile != stdin)
+			fclose(infile);
+
+		/* Finished! */
+		return (rc);
```
cperciva: Is this what we want here?  If `rc` is nonzero, shouldn't we print one of the error messages and `exit(1)`?

gperciva: Added a commit for REBASEing.

cperciva:
LGTM

gperciva:
Rebased, ready for merge.

PR #194, opened 2019-07-11
memlimit: only use RLIMIT_DATA if neither _AS nor _RSS are present

gperciva:


gperciva:
This fixes `make test` on NetBSD.

I'm really uncertain whether we should drop `RLIMIT_DATA` entirely, or keep it as a fallback if the system doesn't have `RLIMIT_AS` nor `RLIMIT_RSS`.

cperciva:
Can you make this conditional on HAVE_MMAP ?  That way we won't be breaking the intention of `RLIMIT_DATA`.  (And also continue to use `RLIMIT_DATA` if we don't have `_AS` or `_RSS`.)

cperciva:
#196 is better.

PR #195, opened 2019-08-01
scryptenc: refactor with scryptdec_file_load_header()

gperciva:


gperciva:
This commit originated in #193 (which probably won't be merged), but I think this change is worthy of consideration by itself, for readability.

PR #196, opened 2019-08-06
memlimit: don't use RLIMIT_DATA if we HAVE_MMAP

gperciva:


gperciva:
This is a revised version of #194.  Normally I'd just change the commit in the earlier PR, but in this case I thought it would be good to have a meaningful branch name (when looking at the git history), and we can't change the branch name that a PR is based on.

cperciva:
I like this change, but I think it needs more explanation in the commit message.  Can you rebase with the following (with appropriate line breaks since the github website doesn't understand what "80 characters" means)?

```
The memlimit code serves two purposes:
1. Making sure that we don't try to use more memory than we can.
2. Trying to avoid using more memory than we should.

On systems without mmap, scrypt uses malloc and RLIMIT_DATA is likely to be a hard limit on what we can allocate; so in that case we respect the value.

On systems with mmap, RLIMIT_DATA has historically still been a useful indication of how much memory "well behaved" programs should use; but as recent mallocs have switched from using sbrk to using mmap(MAP_ANON) it has become common for malloc to not respect RLIMIT_DATA.  As a result, there are now systems where RLIMIT_DATA has "fallen behind" -- with scrypt being the only casualty, since nobody else ever checks the value explicitly.

By looking at RLIMIT_DATA only in the !HAVE_MMAP case, we respect it when it is important while ignoring it in the cases where it is liable to be meaningless.
```

gperciva:
Rebased PR with your commit message.

PR #198, opened 2019-08-14
Cannot combine -P with infile=stdin

gperciva:


main.c
```diff
@@ -162,6 +162,12 @@ main(int argc, char *argv[])
 		}
 	} else {
 		infile = stdin;
+
+		/* Error if given incompatible options. */
+		if (devtty == 0) {
+			warn0("Cannot combine -P and infile=-");
```
gperciva: This would look less "punctuation-y" if we wrote "infile=stdin", but what the user actually would have written on the command-line is "`scrypt dec -P -`".

gperciva:
I'm 95% confident that it's true that we can't combine them, but I might be wrong.

Even if we can't combine -P and infile=stdin, it's still possible to automate scrypt using tools like expect.  Namely, the passphrase can be given automatically without -P, as long as there's a script or program that monitors the output for "Please enter passphrase:".

main.c
cperciva: How about "Cannot read both passphrase and input file from standard input"?

scrypt.1
```diff
@@ -89,7 +89,7 @@ Use at most
 seconds of CPU time to compute the derived encryption key.
 .It Fl P
 Always read passphrase from stdin, and do so only once even
-when encrypting.
+when encrypting.  This cannot be used if infile is also stdin.
```
cperciva: I'd make that "stdin (aka. '-')".

gperciva:
Updated with rebases, since the commits are short enough.

PR #205, opened 2019-09-20
scrypt info: specify that memory is for decrypting

gperciva:


gperciva:
Example:
```
$ ./scrypt info fast 
Parameters used: N = 1024; r = 8; p = 1;
    Decrypting this file requires at least 1.0 MB bytes of memory
```

gperciva:
I ran `scrypt info FILENAME` while testing libscrypt stuff, and was puzzled by the message for 5 seconds.  Since I was the one who added `info`, I figured this was a really bad sign about the output.

PR #206, opened 2019-09-20
build: use libscrypt-kdf in main binary

gperciva:
This dynamically links the scrypt binary to libscrypt-kdf if it's being
generated.

(A completely statically linked scrypt is not desirable -- we use libcrypto,
so we need to be able to replace it easily for security fixes.)

It might seem odd that if we're using libscrypt-kdf, we still include a
number of files in scrypt_SOURCES.  That's because we restricted the
symbols in libscrypt-kdf via:

    libscrypt_kdf_la_LDFLAGS=       -version-info 1                 \
        -export-symbols-regex 'crypto_scrypt$$'

This is because the "public API" of libscrypt-kdf is only a single
function -- scrypt_kdf(), which is `#define`'d as crypto_scrypt -- and we
don't want people starting to rely on any private APIs.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -28,7 +28,6 @@ crypto_scrypt_files=	lib/crypto/crypto_scrypt.c			\
 			libcperciva/util/warnp.h
 
 scrypt_SOURCES=		main.c					\
-			$(crypto_scrypt_files)			\
 			lib/scryptenc/scryptenc.c		\
 			lib/scryptenc/scryptenc.h		\
 			lib/scryptenc/scryptenc_cpuperf.c	\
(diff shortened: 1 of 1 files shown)


gperciva:
This is a starting point for discussion.

(I suspect that this work will never make it into scrypt, but I don't know enough about libraries and low-level POSIX systems to argue whether this is a good idea or not.)

barak:
> This is because the "public API" of libscrypt-kdf is only a single
> function -- scrypt_kdf(), which is #define'd as crypto_scrypt -- and we
> don't want people starting to rely on any private APIs.

Perhaps this should be viewed as an opportunity for a slight refactoring?

gperciva:
Are you referring to the `#define`?  We thought that if the library was called scrypt-kdf, it would make sense for the (only) function to be `scrypt_kdf()`.  But internally, the actual file is `crypto_scrypt.c` because it's crypto code that implements the scrypt function; that makes a lot of sense to us and I can't see it changing.

cperciva:
Not really a fan.  If we were unconditionally building/installing the shared object this might be ok, but I *really* don't like the idea that the scrypt binary will build differently depending on whether an optional component of the package is being installed or not.

Considering how small the amount of code we're duplicating is, I don't see much point anyway...

gperciva:
Just to put some numbers on the compiled sizes:

condition | gcc 7.4.0 | clang 8.0.0
--- | --- | ---
scrypt binary without libscrypt-kdf | **290k** | **190k**
--- | --- | ---
scrypt binary linked to libscrypt-kdf | 258k | 160k
libscrypt-kdf | 123k | 82k
total | **380k** | **242k**

(This is comparing a plain old `make` on Ubuntu 18.04 (so gcc 7.4.0) vs. my tweaked clang with `-Weverything` and disabling errors as necessary.  Both compilations included `-O2 -g` and no attempt was made to tweak compiler options to reduce size.)

PS: compiling it with `clang -Os`, without `-g`, and stripping, produces a **48k** binary.  So if somebody's working on an embedded system with very limited storage, these sizes can be reduced considerably.

gperciva:
This is something that I've been wondering about for a few months, so I'm glad that I finally investigated it in more detail.  I also think that it's worth keeping this branch around, in case it's useful for somebody (e.g,. including it on picoBSD?).

PR #209, opened 2019-10-12
keep the comment style uniform

Alanscut:

PR #210, opened 2019-10-15
Add STYLE document

gperciva:


gperciva:
This also adds it to the `Makefile.am` and `release-tools/mktarball.sh`.

release-tools/mktarball.sh
```diff
@@ -10,7 +10,7 @@ RELEASEDATE=`date "+%B %d, %Y"`
 
 # Copy bits in
 mkdir ${DESTDIR} ${DESTDIR}/autotools
-cp main.c FORMAT COPYRIGHT BUILDING ${DESTDIR}
+cp main.c FORMAT COPYRIGHT BUILDING STYLE ${DESTDIR}
```
cperciva: Hmm, do we need to ship STYLE?  Do we expect anyone to send us a patch after looking at source code from the tarball rather than looking at the git repository?

gperciva: Hmm, we're inconsistent elsewhere.  `spiped` and `kivaloo` include STYLE, but `tarsnap` doesn't.  I wouldn't say that we need to ship it, but I also wouldn't say that we don't need to ship it.

Answering a slightly different question: I can't imagine anybody complaining about us including it, but if we don't include it then we might get a helpful bug report about that "oversight".  And my recent interest in this arose from somebody who very kindly produced a PR for scrypt, but which didn't match the (non-existent) `STYLE` (#209).

So I'd lean towards shipping it, but if you're confident that we don't need it, I can take it out of the tarballs (here and elsewhere).

cperciva: Yeah, fair enough.  It's small enough that it wouldn't get in the way any more than the other "documentation" files.

PR #214, opened 2019-11-03
define scryptenc error states

gperciva:


gperciva:
Unless you have `scryptenc.h` memorized, it's hard to judge whether code like this is correct:
```
	/* Encrypt data. */
	if ((key_enc_exp = crypto_aes_key_expand(key_enc, 32)) == NULL) {
		rc = 5;
		goto err1;
	}
	if ((AES = crypto_aesctr_init(key_enc_exp, 0)) == NULL) {
		crypto_aes_key_free(key_enc_exp);
		rc = 6;
		goto err1;
	}
```

"Does 6 mean 'out of memory', or should that be 7?  And what on earth did 5 mean again?"

It's much easier to read if we replace those magic values with things like `SCRYPTENC_EOPENESSL` and `SCRYPTENC_ENOMEM`.

Currently a draft PR; I'm anticipating a few renamings of the `#define` names I came up with.

lib/scryptenc/scryptenc_cpuperf.c
```diff
@@ -34,6 +34,7 @@
 #include "crypto_scrypt.h"
 #include "monoclock.h"
 
+#include "scryptenc.h"
```
gperciva: I considered moving all the `#define` lines into a new `scryptenc_errno.h` to avoid having to add this line, but I couldn't think of a sensible case in which somebody  would want to include `scryptenc.h` but not have access to the error definitions.

lib/scryptenc/scryptenc.h
```diff
@@ -62,23 +62,22 @@
...
- * 7	data is not a valid scrypt-encrypted block
- * 8	unrecognized scrypt format
- * 9	decrypting file would take too much memory
- * 10	decrypting file would take too long
- * 11	password is incorrect
- * 12	error writing output file
- * 13	error reading input file
- */
+
+/* Return codes from scrypt(enc|dec)_(buf|file|prep): */
```
cperciva: Is the indentation weird here, or is it just github being stupid?

lib/scryptenc/scryptenc.h
```diff
@@ -62,23 +62,22 @@
...
+#define SCRYPTENC_OK		0	/* success */
+#define SCRYPTENC_ELIMIT	1	/* getrlimit or sysctrl(hw.usermem) failed */
+#define SCRYPTENC_ECLOCK	2	/* clock_getres or clock_gettime failed */
+#define SCRYPTENC_EKEY		3	/* error computing derived key */
+#define SCRYPTENC_ESALT		4	/* could not read salt from /dev/urandom */
+#define SCRYPTENC_EOPENSSL	5	/* error in OpenSSL */
+#define SCRYPTENC_ENOMEM	6	/* malloc failed */
+#define SCRYPTENC_EDATA		7	/* data is not a valid scrypt-encrypted block */
```
cperciva: I'd probably call this SCRYPTENC_EINVAL.

lib/scryptenc/scryptenc.h
```diff
@@ -62,23 +62,22 @@
...
+#define SCRYPTENC_EVERSION	8	/* unrecognized scrypt format */
+#define SCRYPTENC_EMEMORY	9	/* decrypting file would take too much memroy */
+#define SCRYPTENC_ETIME		10	/* decrypting file would take too long */
```
cperciva: ETOOBIG, ETOOSLOW, maybe?

lib/scryptenc/scryptenc.h
gperciva: I think the indentation looks ok in both the `.h` and in github...?  The previous content had an extra space (because it was a multiline comment), whereas this is a one-line comment.

Some of the later lines do wrap around, such as
```
#define SCRYPT_ELIMIT           1       /* getrlimit or sysctrl(hw.usernamem faile
d */
```
which is distinctly sub-optimal.  Now that you're ok with this in theory, I'll play around with a few ways to attempt to avoid that.

gperciva:
Slight violation of our normal STYLE, but I think it's justified.

Normally we'd require these `#define`s to begin with `SCRYPTENC_`.  However, that would make the line lengths too long and/or awkward.  Namely:

- We definitely want error 5 to include "`_EOPENSSL`" in the `#define`.
- To avoid rewriting the error explanations, the `#define` names must be
  less than 2 tabs-worth (so 15 characters).
- 15 chars minus strlen("`_EOPENSSL`") is 6, which fits "`SCRYPT`" exactly.

Even with the short `SCRYPT_` form, I've still abbreviated `SCRYPT_EWRFILE` and `SCRYPT_ERDFILE`.  I'm not overjoyed at those, but they're still better than the magic 12 and 13.  (Another option would be `SCRYPT_EWRITEF` and `SCRYPT_EREADF`.)

lib/scryptenc/scryptenc_cpuperf.c
cperciva: The normal reason for putting them into `scryptenc_errno.h` would be the opposite -- to allow `scryptenc_cpuperf.c` to get the error definitions without being polluted with the rest of `scryptenc.h`.

But in this case I don't think that the pollution-avoidance benefit would justify having another file, so this is fine.

lib/scryptenc/scryptenc.h
```diff
@@ -62,23 +62,22 @@
...
- * 12	error writing output file
- * 13	error reading input file
- */
+
+/* Return codes from scrypt(enc|dec)_(buf|file|prep). */
+#define SCRYPT_OK	0	/* success */
+#define SCRYPT_ELIMIT	1	/* getrlimit or sysctrl(hw.usermem) failed */
+#define SCRYPT_ECLOCK	2	/* clock_getres or clock_gettime failed */
+#define SCRYPT_EKEY	3	/* error computing derived key */
+#define SCRYPT_ESALT	4	/* could not read salt from /dev/urandom */
```
cperciva, line 71: Can you drop the "from /dev/urandom" from here?  Theoretically we might use a different entropy backend in the future.

cperciva:
Looks good aside from that one comment (which I know isn't new but let's fix it while we're here).  Please fix that and rebase, and then I'll merge.

gperciva:
Rebased onto master, ready for merge.

Issue #215, opened 2019-11-04
Password from stdin alternatives.


masarlabs:
Hi,
I have batch script that used to call Scrypt in a pileline in this way:
1.    { echo "$PASS" ; cat file_to_decrypt ; } | scrypt dec -P -
2.    { echo "$PASS" ; cat file_to_encrypt ; } | scrypt enc -P -

With the new version of Scrypt I get the error:
   scrypt: Cannot read both passphrase and input file from standard input

Is there an alternative way to pass the password, for example from an environment variable or from a file?
If not, is it possible enable the old way to pass the password?

gperciva:
Fascinating!  I didn't expect that to work, but it clearly does in scrypt 1.2.1.  Test case:

```
#!/bin/sh
echo "secret" > secret.txt
{ echo "hunter2" ; cat secret.txt; } | ./scrypt enc -P - > secret.enc
```
(assuming that you're in a directory containing an `scrypt` binary which is version 1.2.1, and you're **not** using `csh`.  It works in FreeBSD `sh` and `bash`; I haven't tested `ksh` or `tsch` or `foosh`)

Background: I stopped this from working in 1306fbac331ac61eaffc48fb3cd33c44589acc65, or https://github.com/Tarsnap/scrypt/pull/198.  That happened in August 2019, but unfortunately I can't remember why I thought it was necessary.  It was right before I added `scrypt info`, but I can't see any hint in those commits which might have prompted me to "clean up" the existing `main.c.` in that manner.


(I still don't see the use case for this -- if you already know the `file_to_encrypt`, surely the shell script is easier to write as `echo "$PASS" | scrypt enc -P file_to_encrypt`.)

If @cperciva thinks that we should allow this behaviour, it'd be trivial to revert those commits and add it to our test suite.

gperciva:
Update: the test case works in `csh` if we replace `{}` with `()`:

```
#!/usr/bin/env csh
echo "secret" > secret.txt
( echo "hunter2" ; cat secret.txt; ) | ./scrypt enc -P - > secret.enc
```

I'm still curious about the goal of having both the password and input file on stdin, though.

gperciva:
A silly example:

```
#!/bin/sh
printf "hunter2\nI like fluffy kittens\n" > secret.txt
cat secret.txt | ./scrypt enc -P - > secret.enc
```

Test it like this:

```
$ ./silly.sh ; scrypt dec secret.enc
Please enter passphrase: *******
[...]

Issue #201, opened 2019-09-17
How to specify passphrase with an automated script


nodesocket:
How can I specify the passphrase with an automated script? Assume I have an ENVAR set `$SCRYPT_PASSPHRASE`.

Tried:

```bash
scrypt enc -P foo.tar foo.tar.scrypt | echo $SCRYPT_PASSPHRASE
```

nodesocket:
Oops.

```
echo $SCRYPT_PASSPHRASE | scrypt enc -P foo.tar foo.tar.scrypt
```

Would be awesome to update the man page with an example of passing the passphrase like this.

Issue #183, opened 2019-06-21
FeatureRequest: Enable scrypt to get password from file or environment variable


dkmbw:
Hi,
i would like to use scrypt in a CICD pipeline. and the current mechanism would need stuff like expect. Is there a possibility, that a password can be used from the process environment or from a certain file which name can be shown via parameter to scrypt?

That would be really helpful.

gperciva:
Can't this be done with -P ?  Something like (untested)

```
cat `pass.txt` | scrypt enc -P secret.txt secret.enc
```

gperciva:
We do this in our test suite: see `tests/02-decrypt-reference-file.sh` for example.  There we use `echo`:

```
echo ${password} | scrypt dec -P $encrypted_reference_file $decrypted_reference_file
```

dkmbw:
though that is indeed a possibility i could sleep sounder if scrypt would care about fetching the password.  there are in personal and professional environments multitudes of failures. For me, a file (given via parameter) or environment variable (name fixed on compilation) is more sane than echoing the password into scrypt. 

current examples on my mind are: 

- Broken Pipes do not encrypt files now
- set -xv stuff accidentally revealing it in logs
- LC/Locales/encoding scrambles the password on echoing 

And i must admit i oversaw the second sentence in the manpage for -P which said give password only ONCE. i thought i had to input it two times via stdin, therefore i thought of expect.

gperciva:
Well, we can avoid `echo`.  For example,

```
$ cd ~/src/scrypt/tests/verify-strings
$ echo "hunter2" > password.txt
$ scrypt dec -P test_scrypt_good.enc output.txt < password.txt 
```

Some shells will let you do

```
$ scrypt dec -P test_scrypt_good.enc output.txt <<< ${PASS}
```
but that's not standard POSIX.

Both of these options appear to be good against `set -xv`.  I'm not certain what you mean by "broken pipes do not encrypt files now"; unfortunately I know relatively little about shells.  But if you have an unencrypted file containing a password, or if you've set it as an environment variable, that seems easier than a pipe for an attacker.

Could you please elaborate on the threat(s) that you anticipate?

dkmbw:
Of course, thank you! 
First i want to adress the Broken Pipes Example:
Let's consider the following code snippet from a potential script in a CICD pipeline:

`
#!/bin/bash
# decrypt password.txt with masterpassword, for example:
ansible-vault decrypt password.txt
# encrypt the files
[...]

PR #216, opened 2019-11-04
Add --passphrase env:VAR

gperciva:


gperciva:
We've had a bunch of requests for non-stdin password input recenty (https://github.com/Tarsnap/scrypt/issues/183, https://github.com/Tarsnap/scrypt/issues/201, https://github.com/Tarsnap/scrypt/issues/215).  I'm not convinced that this is a good idea, but I'm also not convinced that it **isn't** a good idea.

Securely reading a passphrase from a file might be a bit messy, but reading from an environment variable looks like a simple `getenv()` call.  And if the passphrase is in the environment, presumably security isn't a huge deal.

Draft PR for comments.  The real PR would add it to `usage()`, the `README.md`, etc.

main.c
```diff
@@ -118,6 +119,17 @@ main(int argc, char *argv[])
 				exit(1);
 			}
 			break;
+		GETOPT_OPT("--passphrase-environ"):
+			if ((passwd_environ = getenv("SCRYPT_PASSPHRASE"))
```
gperciva: On second thought, we might as well make this `GETOPT_OPTARG` and read from `optarg` here.

gperciva:
I also have a branch with a proof-of-concept `--password-file`, in case you'd like to see that for comparison.

main.c
```diff
@@ -188,7 +194,18 @@ main(int argc, char *argv[])
...
 
 	/* Get the password. */
-	if (0) {
+	if (passwd_env_arg != NULL) {
+		/* We're not allowed to modify the output of getenv(). */
+		if ((passwd_env = getenv(passwd_env_arg)) == NULL) {
+			warn0("Failed to read from ${%s}", passwd_env_arg);
+			goto err1;
+		}
+
```
gperciva: Should we do any sanity testing here?  for example, if somebody accidentally points to an empty environment variable like this?

```
FOO="" ./scrypt enc --passphrase-env FOO secret.txt secret.enc
```

I'm tempted to insist that the password be at least 1 character when reading from `--passphrase-env`, just to make sure that they didn't accidentally give the wrong `$VAR` or that their system didn't default to an empty string for a missing environment variable.  But that would go against the unix / C principle of "allow people to shoot their own feet".

cperciva: I'm inclined to allow zero-character environment variables.  Any system which silently converts an undefined variable to an empty one is going to run into all sorts of issues; for example, GNU's POSIXLY_CORRECT behaviours kick in even if the variable is defined to be empty.

Also, there's probably going to be scripts which use scrypt to do things and users who say "I don't actually *want* this encrypted".

cperciva:
I'm thinking that it might be worth (somewhat) following OpenSSL's example here:

```
       env:var
           Obtain the password from the environment variable var. Since the
           environment of other processes is visible on certain platforms
           (e.g. ps under certain Unix OSes) this option should be used with
           caution.

       file:pathname
           The first line of pathname is the password. If the same pathname
           argument is supplied to -passin and -passout arguments then the
           first line will be used for the input password and the next line
           for the output password. pathname need not refer to a regular file:
           it could for example refer to a device or named pipe.
```

In the case of `openssl enc` the option name is `-pass` but I would make this a proper long option of course.

OpenSSL also supports `-pass pass:password` (which is horrible and we should not do), and `-pass fd:number`, which we might want to do at some point (but I'd wait until someone needs it).

gperciva:
If you really want to follow their example, should we also include
```
pass:password
```
?

gperciva:
sorry, still waking up.  Ignore that last comment.

gperciva:
Closing this approach, as we'll include this as `--passphrase env:VAR` once #221 is in.

main.c
```diff
@@ -116,6 +121,7 @@ main(int argc, char *argv[])
 	struct scryptdec_file_cookie * C = NULL;
 	enum passphrase_entry passphrase_entry = PASSPHRASE_UNSET;
 	const char * passphrase_arg;
+	char * passwd_env;
```
cperciva, line 124: Even though `getenv` returns a `char *`, I would make this `const char *` in order to enforce the "can't be modified" constraint.

cperciva:
Looks good.

gperciva:
Ready for merge.

PR #217, opened 2019-11-05
Alphabetical order for options, add -v to man

gperciva:


gperciva:
I could have sworn that I checked the man-page before 1.3.0, but evidently I missed the fact that we hadn't added `-v`.

PR #219, opened 2019-11-16
Remove extra \n in warn0

gperciva:


gperciva:
... it's just possible that this was deliberate.  When I first saw it, I thought that I'd screwed up `fprintf()` vs. `warn0()` syntax, but consider:

```
$ ./scrypt info --blah
scrypt: illegal option -- --blah

usage: scrypt {enc | dec | info} [-f] [-M maxmem] [-m maxmemfrac]
              [-P] [-t maxtime] [-v] infile [outfile]
       scrypt --version
$
```

vs.

```
$ ./scrypt info --blah
scrypt: illegal option -- --blah
usage: scrypt {enc | dec | info} [-f] [-M maxmem] [-m maxmemfrac]
              [-P] [-t maxtime] [-v] infile [outfile]
       scrypt --version
$
```

I find the second one slightly easier to parse.  That said, the code looks like an error, so I'd rather clean that up.  But I'm biased since I look at the scrypt code way more often than I actually run `scrypt`.  :)

cperciva:
You're right, `warn*` function should never have `\n` in their message strings.

PR #220, opened 2019-11-19
Passphrase method refactor

gperciva:


main.c
```diff
@@ -135,6 +135,10 @@ main(int argc, char *argv[])
 			verbose = 1;
 			break;
 		GETOPT_OPT("-P"):
+			if (passphrase_entry != PASSPHRASE_UNSET) {
```
gperciva: This is technically a change of behaviour, rather than a refactoring.  But I think it's small enough to not need a separate PR.

gperciva:
You can see what this is aiming towards in https://github.com/Tarsnap/scrypt/pull/221.

PR #221, opened 2019-11-19
Add --passphrase method:arg

gperciva:


gperciva:
Depends on #220.

main.c
```diff
@@ -52,12 +52,52 @@ usage(void)
...
 	exit(1);
 }
 
+static int
+parse_passphrase_arg(const char * arg,
+    enum passphrase_entry * passphrase_entry_p, const char ** passphrase_arg_p)
+{
+	size_t passphrase_sep;
+
+	/* Find the separator in "method:arg". */
```
cperciva, line 68: This works but is unusual.  The normal way of doing this would be

```
const char * p;

/* Find the separator in "method:arg" (and fail if there isn't one). */
if ((p = strchr(arg, ':')) == NULL)
    goto err1;

/* Extract the "arg" part. */
*passphrase_arg_p = &p[1];
```

Another option would be to `strdup` and then use `strtok` which exists for this exact purpose.

gperciva: ... and then do I add

```
size_t method_str_len;
method_str_len = p - arg;
if (strncmp(arg, "dev", method_str_len) == 0) {
```

?  it seemed to me that `strcspn` was a nice way to essentially get two strings out of "foo:bar" without extra memory allocation or weird pointer math [1].

FWIW, `strcspn` is approach that we use in `libcperciva/util/readpass.c` for `string_extract_user_group()` for "username", ":groupname", or "username:groupname".  So I should change that as well if there's a different preferred method.

[1] ok, this isn't very "weird" as far as pointer math goes.  But it still seems like something I need to pay attention to.

cperciva: For that, I would just do `if (strncmp(arg, "dev:", 4) == 0) {`.

The code in `setuidgid.c` is different in that we need to match several options (user, :group, or user:group) so I think it makes sense to use strcspn and strlen there.

gperciva: hmm.  Seems like it involves a bit of extra manual effort (when we add "env:" and "file:"), but ok.

gperciva:
added rebase commit

cperciva:
Looks good.

gperciva:
Rebased, ready for merge.

PR #228, opened 2020-04-17
Clarify scryptenc

gperciva:


lib/scryptenc/scryptenc.c
```diff
@@ -135,6 +135,11 @@ pickparams(size_t maxmem, double maxmemfrac, double maxtime,
 		maxN = opslimit / (*r * 4);
 		for (*logN = 1; *logN < 63; *logN += 1) {
 			checkN = (uint64_t)(1) << *logN;
+
+			/*
+			 * We compare against maxN / 2 because logN is one
```
gperciva, line 140: In retrospect this seems obvious, but I think the comment wouldn't hurt because I'm bound to forget about it and then get puzzled again in a year or two.

cperciva:
This is far more confusing to me than the original code without any comment. What does "one power of two higher than the limit would indicate" mean?

If you want to clarify this code, how about a comment along the lines of "Find the largest power of two less than or equal to maxN, which is also the least power of two greater than maxN/2." ?

gperciva:
Revised comment, and made `checkN` a uint64_t as mentioned in https://github.com/Tarsnap/scrypt/pull/229#discussion_r410573815.

PR #229, opened 2020-04-17
Fix warnings

gperciva:


gperciva:
More clang-10 warnings; all about implicit vs. explicit casts to double.

This PR includes #228; up to you whether you want to merge that one first, or do it all at once with this one.

lib/scryptenc/scryptenc.c
```diff
@@ -134,7 +134,7 @@ pickparams(size_t maxmem, double maxmemfrac, double maxtime,
 		*p = 1;
 		maxN = opslimit / (*r * 4);
 		for (*logN = 1; *logN < 63; *logN += 1) {
-			checkN = (uint64_t)(1) << *logN;
+			checkN = (double)((uint64_t)(1) << *logN);
```
cperciva, line 137: Why are we casting an integer to a floating-point type again?

gperciva: `maxN` is a double, because `opslimit` is a double, because `opps` and `maxtime` are doubles.

That said, I now think it makes more sense to make `checkN` a uint64_t, then have
```
if ((double)checkN > maxN / 2)
```

(Another option would be to change `opslimit` to an integer type, but I assume that you deliberately didn't do that.)

gperciva:
Revised PR to accommodate the updated #228.

PR #234, opened 2020-05-08
tests: check that we don't have output on stderr

gperciva:


gperciva:
The first step is acknowledging that there's a problem.  (re Solaris and getopt)

(I'm working on `siglongjmp` tests for libcperciva, but that's hampered by differences between compilers and architectures and optimization flags.  It's still valuable to test the exact binary that was compiled, in case there's any gaps in the upcoming libcperciva tests.)

PR #235, opened 2020-05-12
Configure Solaris

gperciva:


gperciva:
This fixes the POSIX compatibility detection on Solaris 11.4.  No, I don't have a clue why `sh` fails but `${SHELL}` works.

Draft PR, seeking thought about the general style.
- the `case` seems a bit too platform-specific for `configure.ac`.  Tarsnap has an `actarsnap.m4`, so I'm wondering if we should add an `acscrypt.m4` to contain that hack.  (Or maybe you'd prefer to keep this hack more visible?)

- instead of the whole command being inside the `case` statement, I could define a `USE_SH` variable, then do `command -p ${USE_SH}`.  That might make the code more readable.
(I just checked, and this does work.)

PR #236, opened 2020-05-13
libcperciva-import

gperciva:


gperciva:
Needs a minor tweak to `.travis-cl`, this isn't an actual problem. But by chance, I discovered a real problem which is now https://github.com/Tarsnap/libcperciva/pull/245.

gperciva:
This should be good to merge now.

PR #237, opened 2020-05-13
Passphrase /dev/tty once

gperciva:


gperciva:
This is similar to `--passphrase dev:stdin-once`, which is the same as the old `-P` option.

PR #238, opened 2020-05-14
Fix RLIMIT_DATA on OpenBSD

gperciva:


lib/util/memlimit.c
```diff
@@ -63,6 +63,11 @@
 #undef HW_USERMEM
 #endif
 
+/* Is RLIMIT_DATA meaningful on this platform? */
```
cperciva, line 66: I would s/meaningful/relevant/.  There may be systems where malloc invokes sbrk and RLIMIT_DATA is *meaningful* but not *relevant* to us because we're using mmap.

gperciva:
Fixed and rebased, ready for merge.

PR #239, opened 2020-05-17
Separate "find a POSIX sh" from "run libcperciva/POSIX"

gperciva:


gperciva:
Once these two commits are in, there's only a final trivial commit to fix the build on Solaris: https://github.com/Tarsnap/scrypt/pull/240.

gperciva:
Yes, I plan to move a few more lines into `acscrypt.m4`; but that would happen in a separate PR (as a cosmetic change), rather than a "fix build on Solaris" thing.

cperciva:
Is this relevant at all for !Solaris?  I'm struggling to imagine how, without that context, anyone would look at this and say anything other than "why are we adding unnecessary complexity?"

gperciva:
A few reasons for non-Solaris:
- why run a config step (`command -v -p sh`) twice?
- what happens if that step fails?  Wouldn't it be nice to have an error message?
- a fair number of configure steps print output, such as

      checking for a BSD-compatible install... /usr/bin/install -c
      checking for a working dd... /bin/dd
      checking for a sed that does not truncate output... /usr/bin/sed

  Adding a line for

      checking for a POSIX shell... /bin/sh

  is entirely in line with those.

Now, you might be thinking "oh, surely we can always get a POSIX shell!".  But 1) we *know* that's not true (*cough* Solaris), and 2) the whole point of `configure` is to handle weird edge cases.  Will it work on windows subsystem for linux?  Somebody's manually-compiled "linux from scratch"?  Android?  BeOS?

If `command -v -p sh` fails, then I want an error message about just that line.   Otherwise I might end up wasting more hours looking at `libcperciva/POSIX/*.sh` when there's actually no problem in those scripts because I couldn't believe that `command` might work at the beginning of `configure.ac` but fail halfway through.

gperciva:
PS: Yes, if you just look at

```
+POSIX_SH=`command -v -p sh`
 AC_MSG_NOTICE([checking POSIX compatibility...])
 LIBCPERCIVA_DIR=${srcdir}/libcperciva
-LDADD_POSIX=`export CC="${CC}"; command -p sh ${LIBCPERCIVA_DIR}/POSIX/posix-l.sh "$PATH"`
-CFLAGS_POSIX=`export CC="${CC}"; command -p sh ${LIBCPERCIVA_DIR}/POSIX/posix-cflags.sh "$PATH"`
+LDADD_POSIX=`export CC="${CC}"; ${POSIX_SH} ${LIBCPERCIVA_DIR}/POSIX/posix-l.sh "$PATH"`
+CFLAGS_POSIX=`export CC="${CC}"; ${POSIX_SH} ${LIBCPERCIVA_DIR}/POSIX/posix-cflags.sh "$PATH"`
```

then it looks like unnecessary complication.  I'm arguing that the following commit, which adds `CHECK_POSIX_SH`, is a much more autoconf-esque solution.  (And the `AC_MSG_ERROR` in that macro would have saved Tarsnap Backup Inc. quite a bit in wages.)

PR #240, opened 2020-05-17
Fix solaris build

gperciva:


gperciva:
Yes, this needs more documentation about why it's needed.  I put this online in case you wanted to glance at e90d9ab while evaluating https://github.com/Tarsnap/scrypt/pull/239.

Issue #241, opened 2020-05-19
POSIX issue 7 (2017) now required


gperciva:
The recently-added `CHECK_POSIX_SH` [1] uses `command -v`.  This is widely accepted as the proper answer to "how do I portably check if a command exists" [2], and has been part of our `make test` since 2016-04-23 [3]. 

However, `-v` only became mandatory in issue 7; previously, it was a [UP] User Portability Utilities option [4].  The debian `posh` shell does not support it, for example.   Solaris is inconsistent between the shell built-in `command`, `/usr/bin/command`, the man-page, and `/usr/xpg4/bin/command` (the latter one is fully compatible with POSIX issue 7, as far as I can tell).

I'm still investigating, but thought that I should document the issue.

[1] https://github.com/Tarsnap/scrypt/commit/abb7241dd6f091813c6a7b3007ec89422450f176
[2] https://stackoverflow.com/a/677212
[3] https://github.com/Tarsnap/scrypt/commit/4dc4eeb03027c9cf2c6f59f7ec3070b25b7eb383
[4] https://pubs.opengroup.org/onlinepubs/009695399/utilities/command.html

gperciva:
Resolved in #242.

PR #242, opened 2020-05-19
build: avoid POSIX-issue-7-required command

gperciva:


gperciva:
Some discussion in #241.  This fixes the build issue on Solaris, so we won't be required to move `CHECK_POSIX_SH` to the beginning of `configure.ac` "for some reason".

Issue #200, opened 2019-08-19
Feature request: Specify N, r, p in the command line


mihao:
There is a lot of information about what scrypt parameters are recommended at a given time for a given level of security, but there is no easy way to specify these parameters in the command line. I can specify maxmem and/or maxtime, but I won't know how safe my data is going to be, because the values of N and p depend on the state of my machine and they even change from one invocation to another (which is very confusing, by the way). Moreover, decryption may happen on a completely different machine, so maxmem and maxtime have limited usefulness anyway.

 It would be great if we could simply invoke scrypt like this:
`scrypt enc -N1048576 -r8 -p1 secret_file secret_file.scrypt`
Or perhaps this notation would be a bit more convenient:
`scrypt enc -lgN 20 -r 8 -p 1 secret_file secret_file.scrypt`

cperciva:
I don't think it makes sense to specify `r` at the command line; the default value of `8` has been appropriate on a wide range of systems for the past decade and there's no reason to think that it will cease to be appropriate in the foreseeable future.

But setting `N` and `p` via the command line seems like it would be a useful feature.

gperciva:
Since `N` must be a power of two, and passing large numbers on the command-line is annoying, I think it makes sense to pass `--logN` rather than `-N`.  (But it's not a firm belief, so I'm happy to be overruled.)

gperciva:
This was added in #249.

PR #249, opened 2020-06-03
Manually specify scryptenc parameters

gperciva:


gperciva:
This allows us to specify logN, r, p on the command line, as per #200.

I plan to introduce this as 2 or 3 separate PRs; the first of which is #250.

gperciva:
Ready for review.

Big question mark about bounds on `--logN`, `-r`, `-p`.  I think it's reasonable to have some (in addition the hard limits of the scrypt format), but I'm wildly guessing at what would be appropriate.

main.c
```diff
@@ -62,6 +62,21 @@ usage(void)
 	exit(1);
 }
 
+#define GETOPT_PARSENUM_WITHIN(var, min, max)				\
```
cperciva, line 65: We normally write these as
```
#define foo(bar) do { \
        blah blah blah; \
} while (0)
```

cperciva: That is to say, with the `do {` on the opening line and with a level less indentation than you have here.

main.c
```diff
@@ -160,6 +175,9 @@ main(int argc, char *argv[])
 		GETOPT_OPT("-f"):
 			force_resources = 1;
 			break;
+		GETOPT_OPTARG("--logN"):
+			GETOPT_PARSENUM_WITHIN(&params.logN, 10, 32);
```
cperciva, line 179: I'd be inclined to make this range 10--40, just for future-proofing.

We're applying the "make sure we have a reasonable amount of memory/time" tests (assuming no `-f`) right?  So there should already be a seatbelt to protect against crazy parameters.

gperciva: No, there's no "reasonable" checks.  `scryptenc_setup:`

```
        /* Determine parameters. */
        if (P->logN != 0) {
                /* Check logN, r, p, without any resource limits. */
                if ((rc = checkparams(0, 0, 0, P->logN, P->r, P->p,
                    verbose, 1)) != 0) {
```

The idea is that if somebody explicitly specifies N, r, p, then we assume that they know what they're doing.

(this change was in a previous PR, in case you're wondering why you didn't see that code in this PR)

gperciva: * the "without any resource limits" refers to the final `1` in `checkparams()`, which enables `force`.

gperciva:
Rebased, ready for review.

main.c
```diff
@@ -54,14 +54,26 @@ usage(void)
...
+	    "              [--passphrase method:arg] infile [outfile]\n"
 	    "       scrypt --version\n");
 	exit(1);
 }
 
+/* Parse a numeric optarg within a GETOPT context.  (Requires ch and optarg.) */
+#define GETOPT_PARSENUM_WITHIN_UNSIGNED(var, min, max) do {		\
+	if (PARSENUM((var), optarg, (min), (max))) {			\
+		if (errno == ERANGE) {					\
+			warn0("%s must be between %zu and %zu"		\
```
cperciva, line 69: You're printing these as `size_t` variables, but you're not casting them to that type.

gperciva: Argh.

I'm tempted to do a whole parsenum-style "print as double, intmax_t, or uintmax_t" macro, but I can't find enough instances of "parse a number within this range" in our code to make it IMO worth it.

Since unattached numbers are intepreted as `int`, I renamed this to `GETOPT_PARSENUM_WITHIN_INT`.

cperciva: Well, I guess you could do that.  Although I would still prefer to see explicit casts inside the printf.

But I was more thinking of casting to `uintmax_t` and using `%ju`...

gperciva: That was my first thought, but then I figured that you'd say it was over-engineered.

... ok, updated.

cperciva:
In general, macros *should* be over-engineered, since they're inherently somewhat opaque; it's easy for someone using one in the future to not be aware of their limitations.

LGTM, please rebase.

gperciva:
Rebased.

I also fixed a few other things:
- renamed the test to `09-explicit-params.sh` (it previously had `manual` instead of `explicit`)
- fixed the alphabetical order of `[-p value]` in the man page.
- tweaked the man-page option text to say:

  > If such explicit parameters are given, the resource limits set by -M, -m, and -t are not enforced.

  (Previously it just said that -M, -m, and -t were ignored)

cperciva:
What happens if someone specifies `--logN 15` but does not specify values for *r* and *p*?

gperciva:
Assertion error, as desired?

... oh right, we should have a more polite message in `main.c`.

gperciva:
Added the checks for attempting to partially set explicit params.

At first, I combined it all into two lines (like the assertions), but I found that unpleasantly complicated.  More importantly, I couldn't find a good wording for the error message that would make sense to people who didn't know what was happening without it exceeding 80 characters.

So instead, I split into 3 checks.  That meant the error message could mirror the man-page sentences, and vastly simplified the `if` statements.

cperciva:
LGTM.

gperciva:
Rebased, ready for merge.

PR #250, opened 2020-06-03
scryptenc_params: change public api

gperciva:


gperciva:
This is the first part of #249.  I think it will be easier to add optional N, r, p as command-line parameters if we pass a single `struct scryptenc_params` instead of distinct arguments.

This PR introduces such a struct, but does not introduce any change of behaviour.

PR #253, opened 2020-06-17
scryptenc_params: refactor with exact params in struct

gperciva:


gperciva:
This is the next step towards #249.  This might be too "granular" a change; I originally had more in the "next step" branch, but then I chickened out and decided to only put this much forward, since it has important questions.

1. Should we specify `N`, or `logN`?  I went with `logN`, because that's what the header format calls for, and because it's an error to specify a non-log2 `N`.

2. what data type should `logN` be?  `scryptenc.c` uses `int logN`, which I found slightly puzzling.  The `FORMAT` specifies that it must be between 1 and 63 inclusive, as does the code in `checkparams`, so I would have thought that a `uint8_t` would be more suitable, especially since `r` and `p` are both `uint32_t` and `crypto_scrypt` ultimately wants a `uint64_t N`.

At the moment I went with `int logN` for consistency with the existing `scryptenc.c`.

cperciva:
I normally use `int` for "integer value with an explicit small range which is being used within C code".  So, something like 0--9 or 1--63.  Since `r` and `p` are theoretically limited only by the range of the types they'll be encoded into, it make sense for them to be `uint32_t` from the start.  So yes, specify `int logN`.

I'd like to have a bit more code in here to review -- it's hard to make much of "add unused variables to a structure".

gperciva:
Updated with more code, and changed the PR name to match.  Unfortunately the branch name makes less sense now, but oh well.

lib/scryptenc/scryptenc.h
```diff
@@ -66,6 +66,11 @@ struct scryptenc_params {
 	size_t maxmem;
 	double maxmemfrac;
 	double maxtime;
+
+	/* Precise parameters. */
```
cperciva, line 70: "Precise" seems like the wrong word here.  "Explicit"?  "Manually specified"?

lib/scryptenc/scryptenc.c
```diff
@@ -230,38 +230,35 @@ scryptenc_setup(uint8_t header[96], uint8_t dk[64],
...
-	uint32_t p;
 	SHA256_CTX ctx;
 	uint8_t * key_hmac = &dk[32];
 	HMAC_SHA256_CTX hctx;
 	int rc;
 
 	/* Pick values for N, r, p. */
 	if ((rc = pickparams(P->maxmem, P->maxmemfrac, P->maxtime,
-	    &logN, &r, &p, verbose)) != 0)
+	    &P->logN, &P->r, &P->p, verbose)) != 0)
```
cperciva, line 241: This changes program behaviour, since the values `logN`, `r`, and `p` are now returned to the caller via the structure `P`.

This is fine, but the functions should be documented as returning the chosen parameters via the structure.

lib/scryptenc/scryptenc.h
gperciva: I've used all 3 at various points in history of this feature.  I'll go with "explicit", then.

gperciva:
While I was checking the full set of changes, I rediscovered that the explicit parameters will only come into play in `scryptenc_setup()`, because of course in `scryptdec_setup()` the parameters come from `header[96]`.

I assumed that you'd want the `scryptdec_setup()` to have the same changes so that it's the same (hence why they're included in this PR), but it occurs to me that I should check.

gperciva:
I've (mostly) gone back to my previous idea of PR division: this one adds the explicit parameters to the struct, but is strictly only refactoring.  The "return values via ${params}" changes have been moved to https://github.com/Tarsnap/scrypt/pull/254.  If you want to deal with those changes at the same time as these, we can close this PR and only work on the other one.

Both PRs have many additional function API comments, as well as increased commit messages.

PR #255, opened 2020-06-22
build: fix cpusupport linkage for test_scrypt

gperciva:


gperciva:
Fixes Solaris build with `cc` and `c99`.  Commit message was mostly taken from your email.

PR #259, opened 2020-07-08
scryptenc_params: use explicit in scryptenc_{buf,file}

gperciva:


gperciva:
This is the final PR relating to #249  that changes `scryptenc.{c,h}`.  After this PR, the only changes will be to `main.c`, the man page, and the test suite.

PR #263, opened 2020-07-19
Don't assume `--force` with explicit params

gperciva:


gperciva:
This changes the behaviour of "scryptenc with explicit parameters" to match what you thought it was (in https://github.com/Tarsnap/scrypt/pull/249#discussion_r456950860).

I'm not advocating that this is a good idea, but neither do I claim that it's not a good idea.

cperciva:
Can we make the behaviour "if you specify explicit parameters, you get what you asked for but you might also get a warning"?

gperciva:
Hmm.  So far, `lib/scryptenc/*` has avoided using any `warn0()`; I assume that's why we have all the error codes in `scryptenc.h` which are then interpreted by `main.c` to print error messages.  I figured this was an important part of your sneaky master plan.

OTOH, `scryptenc.c` includes lots of libcperciva files which use `warn0` or `warnp`, so it's not like we can link to `scryptenc.o` without `warnp.o` as well.

I've added a commit which adds the "gives you what you asked for, but prints a warning" functionality.

Other ideas I (semi-seriously) considered:
- pass a callback to `scryptenc`, purely to print a warning (and therefore keep `warnp.h` out of `scryptenc.c`.
- add a `scryptenc_prep()` function which handled the validation of parameters, so that it could return back to `main.c` for warning-printing before calling `scryptenc_file`.

cperciva:
FWIW, the "sneaky master plan" was to keep *scrypt* as minimal as possible in its dependencies, since it's likely to be reused in other projects.  But *scryptenc* is a far more niche application so I'm fine with it using larger chunks of libcperciva.

lib/scryptenc/scryptenc.c
```diff
@@ -241,10 +242,25 @@ scryptenc_setup(uint8_t header[96], uint8_t dk[64],
 		/* Check logN, r, p. */
 		if ((rc = checkparams(P->maxmem, P->maxmemfrac, P->maxtime,
 		    P->logN, P->r, P->p, verbose, force)) != 0) {
+			/* Warn about resource limit, but suppress the error. */
+			if (rc == SCRYPT_ETOOBIG) {
+				warn0("Explicit parameters require more memory "
+				    "than is available; continuing anyway.");
```
cperciva, line 248: Rather than "... continuing anyway", how about "Warning: Explicit parameters may require too much memory" (and similarly for ETOOSLOW below)?  It's possible that someone will use explicit parameters on systems where the memlimit code doesn't work, so I think the "may" is meaningful.

gperciva: If I did
```
$ scrypt enc --STUFF foo bar.enc
scrypt: Warning: Explicit parameters may require too much memory
```

and then nothing happened for a while, I'd assume that the program was in an infinite loop or was waiting for user input for some reason.  So I'd hit <enter> a few times, then ctrl-c.

I'm happy to add the "may" (or "might") in there, but I think we should have some language to indicate that the process is continuing.

gperciva:
> FWIW, the "sneaky master plan" was to keep scrypt as minimal

I assume that you meant `crypto_scrypt` instead of `scrypt`.

gperciva:
New draft of warning messages for time and memory.

cperciva:
I think most users of unix tools understand that programs operate quietly by default.  And they would probably hear their CPU fan turning on as scrypt runs, so I'm not too worried about people thinking that scrypt is just hanging.

gperciva:
As per dfa089d, the messages are now:

    Warning: Explicit parameters might exceed memory limit
    Warning: Explicit parameters might exceed time limit

cperciva:
Yes.  I think this is good.  Please clean up the history and I'll merge it.

gperciva:
Rebased, ready for merge.

PR #271, opened 2020-08-31
libcperciva import and trivial comments

gperciva:


gperciva:
There were only a handful of "add () to function names" changes, so I lumped them in with the libcperciva import.

Issue #277, opened 2020-09-16
Misleading documentation on the scrypt website regarding libscrypt-kdf


sternenseemann:
Seems like the paragraph about "Using scrypt as a KDF" on the [website](https://www.tarsnap.com/scrypt.html) was written before `libscrypt-kdf`: It suggests the header is `crypto_scrypt.h` and the function `crypto_scrypt()`. I believe this should be `scrypt-kdf.h` and `scrypt_kdf()` respectively?

gperciva:
Thanks, @sternenseemann!

Definitely agreed that `scrypt_kdf()` should be mentioned on the website (and README).  I'm checking with @cperciva about whether we want to keep the current discussion about `crypto_scrypt()` or not.  If this was a new project, we'd only mention the `scrypt_kdf()` function, but there's ten years of people using `crypto_scrypt()` because that's all we had in the past.

sternenseemann:
The function name is not that big of an issue in my opinion since they both can be used since `scrypt_kdf` is just an alias for `crypto_scrypt`. My main concern is that the documentation to me seemed to suggest that the header was `crypto/scrypt.h` or `crypto_scrypt.h` while `make install` would only install `scrypt-kdf.h` and likewise the debian package also only installs that header.

gperciva:
Good point, thanks!  I've rewritten it to discuss `libscrypt-kdf` first.

One tiny nitpick: `make install` will only install `scrypt-kdf.h` if you use `./configure --enable-libscrypt-kdf`; it doesn't happen by default.  Granted, it looks like many packagers are enabling this (which is great!), so maybe we don't need to worry so much about people installing it from scratch.

gperciva:
The README and website have been updated, thanks!

PR #278, opened 2020-09-19
README: mention scrypt_kdf()

gperciva:


gperciva:
Partially addresses #277 (if this is accepted, I'll update the website accordingly).

Original discussion about the public header for `libscrypt-kdf` is here:
https://github.com/Tarsnap/scrypt/pull/168#issuecomment-485275514

My first instinct was to remove all mentions of `crypto_scrypt()` from that section, but historically people have used scrypt by copying files into their own project and calling that function.  I don't know if we should try to discourage that now, so I opted for a "minimal" change of mentioning `scrypt_kdf()` in the context of the optional `libscrypt-kdf`.

gperciva:
Updated the PR to mention `scrypt_kdf()` first, and significantly downplay the "copy source files into your project" option.  See it in action:
https://github.com/Tarsnap/scrypt/tree/discuss-scrypt-kdf#using-scrypt-as-a-kdf

PR #281, opened 2020-10-09
passphrase_entry.[c,h]

gperciva:


gperciva:
This prepares for sharing this code with the tarsnap client.

Rather than adding a new `passphrase_entry.h`, I could shove this into `readpass.h` (and perhaps a `readpass_entry.c`).  I'm not certain which would be cleaner, API-wise.

cperciva:
I think this is good.  There are many "readpass" functions in the world and AFAIK they all read passwords from the terminal; I think it makes sense to keep that narrow function as it is and have this as a more general-purpose wrapper.

PR #294, opened 2020-12-30
tests: allow skipping the 1 GB test with SMALLMEM=1

gperciva:


gperciva:
This is primarily intended for testing on EC2 .nano instances, which only give us 512 MB of memory.

AFAIK, the N=1048576 is the only test in all our software that doesn't run comfortably on a nano instance.  (I haven't tested kivaloo on those, though.)

tests/verify-strings/test_scrypt.c
```diff
@@ -24,16 +24,24 @@ main(int argc, char * argv[])
...
-	(void)argv; /* UNUSED */
+	/* Do we want to skip the large-memory test? */
+	if ((argc == 2) && (strcmp(argv[1], "1") == 0))
+		skip_largemem = 1;
+
+	for (test_i = 0; test_i < num_tests; test_i++) {
+		test = &tests[test_i];
+
+		/* Skip the 1 GB memory test (if desired). */
+		if ((skip_largemem) && (test_i == 3))
```
cperciva, line 42: I would have written this as `test == &tests[3]` and avoided the whole `test_i` thing...?

cperciva: Alternatively, added an `int largemem;` field to `struct scrypt_test` and then this could just be `if (skip_largemem && test->largemem)`.

gperciva: Do you really like the original way of looping?  It throws me for a loop (pun intended) whenever I look at that file.

I have to admit that I like the `int largemem` solution, though.

cperciva: I'm not opposed to the original way of looping.  Not sure if I would write it like that now, but I wouldn't change it without a good reason either.

gperciva:
Rebased with `int largemem`.

PR #304, opened 2021-02-10
Libcperciva import

gperciva:


gperciva:
We still need to specify CPUSUPPORT CFLAGS in `Makefile.am` for now.

gperciva:
Oh, I should also mention that this is the final set of PRs for the current round of libcperciva imports.

PR #309, opened 2021-03-23
passphrase_entry: add prompt & confirmprompt as args

gperciva:


gperciva:
This isn't necessary in scrypt itself, but tarsnap has different prompts for different situations.

PR #314, opened 2021-06-13
Libcperciva import

gperciva:


gperciva:
I previously mused about splitting this into two separate PRs, but since there's so few changed lines -- almost everything is in new files -- I figured it could work as a single large PR.

PR #317, opened 2021-08-28
Fixes

gperciva:


gperciva:
These arose from #317, thanks @msimonsson!

Issue #316, opened 2021-08-28
Verbose output is suppressed by warning


msimonsson:
`scrypt enc --logN 20 -p 1 -r 8 -v -- test test.scrypt`

> Please enter passphrase:
> Please confirm passphrase:
> Parameters used: N = 1048576; r = 8; p = 1;
>     Decrypting this file requires at least 1.0 GB bytes of memory (1.1 GB available),
>     and will take approximately 0.3 seconds (limit: 5.0 seconds).

`scrypt enc --logN 21 -p 1 -r 8 -v -- test test.scrypt`

> Please enter passphrase:
> Please confirm passphrase:
> scrypt: Warning: Explicit parameters might exceed memory limit

P.S. There is a typo in the man page for the verbose option `--v` (double-dash) not `-v`.

msimonsson:
Some additional thoughts.

I changed the following before testing:
```diff
% diff scrypt-master/lib/scryptenc/scryptenc.c scrypt/lib/scryptenc/scryptenc.c
203a204,206
>         if (verbose)
>             display_params(logN, r, p, memlimit, opps, maxtime);
>
214d216
<       }
216,217c218,220
<       if (verbose)
<               display_params(logN, r, p, memlimit, opps, maxtime);
---
>         if (verbose)
>             display_params(logN, r, p, memlimit, opps, maxtime);
>       }
```

```diff
% diff scrypt-master/lib-platform/util/memlimit.c scrypt/lib-platform/util/memlimit.c
29a30,31
> #define DEBUG
>
```

OS: FreeBSD 13

1. Only one of the warnings will be shown when both limits are exceeded. `SCRYPT_ETOOBIG` is checked before `SCRYPT_ETOOSLOW`.

2. The memory limit is checked against how much memory scrypt allowed, not available memory.

`echo "foobar" | ./scrypt-test enc --logN 21 -p 1 -r 8 -P -v -- test /dev/null`
> Memory limits are:
>         usermem:        8980013056
>         memsize:        18446744073709551615
>         sysinfo:        18446744073709551615
>         rlimit:         18446744073709551615
>         sysconf:        17098350592
> Allowing up to 1122501632 memory to be used
> Parameters used: N = 2097152; r = 8; p = 1;
>     Decrypting this file requires at least 2.1 GB bytes of memory (1.1 GB available),
>     and will take approximately 0.6 seconds (limit: 5.0 seconds).
> scrypt-head: Warning: Explicit parameters might exceed memory limit

[...]

PR #320, opened 2021-09-02
Clarify memlimit maxmemfrac

gperciva:


gperciva:
While looking into issues mentioned in #316, I found two problems with maxmemfrac.

1. the API comment for `memtouse()` doesn't mention that it treats `maxmemfrac` >0.5 like 0.5, and that a `maxmemfrac` of 0 is also treated as 0.5.

   (this is mentioned in `scryptenc.h`, but not `memlimit.h`)
2. the binary allows users to specify a `maxmemfrac` of up to 1.0, without any indication that it will only use 0.5.

`memlimit.c` was added in 2021-08-18 for scrypt-1.1, and it contained the "silently treat >0.5 as 0.5" thing.  I'm not certain why you wanted that.

This PR is a sketch of a possible solution, but this depends on exactly how you want the library function and binary to operate.

Issue #318, opened 2021-08-28
FTBFS on systems without getconf


ismaell:
Some systems don't provide `getconf`. The `CHECK_POSIX_SH` macro fails on those systems:
https://github.com/Tarsnap/scrypt/blob/c83b315c2ce29719339f56e1b96a28f7f6acf93d/m4/check_posix_sh.m4#L6

It should, at least, behave like a standard macro and allow overriding the vaule from the command line.

gperciva:
I believe that `getconf` is part of POSIX:
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/getconf.html

I can look into allowing people to override it, as a workaround for systems which are not POSIX-compliant.

gperciva:
@ismael: also, which system(s) don't have `getconf`?  I should add a note in the `BUILDING` file about this.

I thought that the FTBFS acronym came from debian & ubuntu, but I can't see any open bug reports in their package trackers.

ismaell:
The systems I'm dealing with right now are:
- Source Mage GNU/Linux, musl-libc variant (also, our default `sh` may vary)
- NuttX

On most systems the test doesn't make much sense because there should be only one `sh` in `PATH`, but on systems that matters, depending on the current `PATH` you may get different versions of `getconf` returning different answers, e.g. in Solaris:
- `/usr/xpg4/bin/getconf`
- `/usr/xpg6/bin/getconf`
- `/usr/bin/getconf`

I think the one in `/usr/bin/` returns the same values as the one in `/usr/xpg4/bin/`...

In any case, should't the compliance be somehow tested? otherwise it's blind faith, and the system could be making bogus promises...

(yes, FTBFS came from Debian, but everyone understands it now and it's short :smiley:)

gperciva:
Huh, NuttX looks cool!

Ok, this is a tricky thing that tests the bounds of POSIX compatibility, and I'm quite willing to believe that I got it wrong!  This whole thing arose because of Solaris, namely #241 (fixed in #239 and #242).

My current understanding is that a POSIX-compliant system:
* `command -p`:
   > Perform the command search using a default value for PATH that is guaranteed to find all of the standard utilities.
  > https://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html

  So it shouldn't matter what `PATH` the user sets.

* `command -p getconf PATH`: retrieves that default PATH.

  In fact, the RATIONALE for `getconf` specifically notes:
[...]

PR #321, opened 2021-09-03
build: allow users to manually specify POSIX_SH

gperciva:


gperciva:
I'm mostly neutral about this PR.  There's a report in #318 that some OSes (which appear to be in early development) do not include `getconf`.

- on one hand, `getconf` is a mandatory part of POSIX, so my first instinct is that those systems should implement it.

- on one hand, the whole point of `configure` is to work around buggy OSes.  For `scrypt`, we might not care so much, but we definitely want to be flexible for `tarsnap`.
 
This allows users to do `./configure POSIX_SH=/my/weird/path/sh`.  It appears in `./configure --help` as:

```
Some influential environment variables:
  CC          C compiler command
...
  POSIX_SH    path of a POSIX shell
```

I haven't added this to `BUILDING`, since I think it's quite unlikely that anybody on a stable OS will need it.

PS: I've spent too long trying to decide whether the `--help` string should say "path of a POSIX shell" or "path to a POSIX shell".

cperciva:
I think this is good.  For the "strict POSIX" tools (spiped, kivaloo) I would say no (and suggest `env PATH=... make`) but for scrypt and tarsnap we want to support more broken platforms.

PR #322, opened 2021-09-03
Add SCRYPT_EBIGSLOW

gperciva:


gperciva:
This allows us to print warnings for both memory and cpu:
```
$ echo "hunter2" | ./scrypt enc --logN 21 -p 1 -r 8 -P -- config.h /dev/null
scrypt: Warning: Explicit parameters might exceed memory limit
scrypt: Warning: Explicit parameters might exceed time limit
$
```

I considered switching the `SCRYPT_EFOO` defines to a bitfield (1 << 1, 1 <<2, 1<<3, etc.) so that we could represent multiple errors by ORing them together, but on closer examination I decided that the only "multiple errors" we might want to represent would be these two, and it didn't make sense to change the whole error-handling infrastructure rather than simply adding another error code.

cperciva:
I get the feeling that the relevant code in scryptenc_setup would be better as a switch statement.  I'll merge this now but when you get a chance see if you can clean it up.

PR #323, opened 2021-09-16
scryptenc: use switch() instead of ifs

gperciva:


lib/scryptenc/scryptenc.c
```diff
@@ -248,28 +248,32 @@ scryptenc_setup(uint8_t header[96], uint8_t dk[64],
 		/* Check logN, r, p. */
 		if ((rc = checkparams(P->maxmem, P->maxmemfrac, P->maxtime,
 		    P->logN, P->r, P->p, verbose, force)) != 0) {
+			switch(rc) {
 			/* Warn about resource limit, but suppress the error. */
```
gperciva, line 252: I think this is what you asked for, but I'm not a huge fan.

Where should the comment be, if it relates to 3 of the 4 cases?  For now, I put it on top of the first of those cases, but I think that's not as clear as the previous "comment, followed by three `if` statements".

lib/scryptenc/scryptenc.c
```diff
@@ -248,28 +248,32 @@ scryptenc_setup(uint8_t header[96], uint8_t dk[64],
...
-			if (rc == SCRYPT_ETOOSLOW) {
+				break;
+			case SCRYPT_ETOOSLOW:
 				warn0("Warning: Explicit parameters"
 				    " might exceed time limit");
 				rc = 0;
-			}
-
+				break;
 			/* Provide a more meaningful error message. */
```
gperciva, line 270: Was the intention to include this in the `switch` (which I did), or leave it separate?

I think that this is sufficiently different -- changing the error message, but not suppressing the error -- that I think it's more clear as its own `if` statement.  On the other hand, it looks really weird to have `switch(rc)` immediately followed by `if (rc)`.

lib/scryptenc/scryptenc.c
```diff
@@ -248,28 +248,32 @@ scryptenc_setup(uint8_t header[96], uint8_t dk[64],
...
-			if (rc == SCRYPT_EINVAL)
+			case SCRYPT_EINVAL:
 				rc = SCRYPT_EPARAM;
+				break;
+			default:
+				break;
```
gperciva, line 275: Stylistically, I'm not certain if you'd prefer a `break` here, or a `/* FALLTHROUGH */`.  I have a sneaky suspicion that, strictly speaking, we might need to have an empty `;` if it's just a comment in the `default`.

cperciva:
Yeah I agree this isn't much of an improvement.

Maybe
```
if (rc == SCRYPT_ETOOBIG || rc == SCRYPT_EBIGSLOW)
    ... one warning ...
if (rc == SCRUPT_ETOOSLOW || rc == SCRYPT_EBIGSLOW)
    ... the other warning ...
if (rc == SCRYPT_ETOOBIG || rc == SCRYPT_ETOOSLOW || rc == SCRYPT_EBIGSLOW)
    rc = 0;
```
would work?  Having the same error messages copied in two places bothers me.

PR #325, opened 2021-09-29
Stricter sanity check for logN

gperciva:


gperciva:
Imported from tarsnap.

PR #328, opened 2021-10-31
Slightly simplify configure with this one weird trick!  (no effect whatsoever)

gperciva:


gperciva:
https://github.com/Tarsnap/tarsnap/pull/501 had a statistically significant improvement with only N=5 tests.  But in this one, there's no difference, even with N=15 tests.

*shrug*

I still think it's worth including, just to keep the build systems in sync.

```
$ ministat times-master.txt times-pr.txt 
x times-master.txt
+ times-pr.txt
+------------------------------------------------------------------------------+
|            x + x   x                                                         |
|            x + x   x     +                                                   |
|+ x x +   + * + * + x x   + * x     +                    + x                 +|
|    |_|_________M__A____A________|__________|                                 |
+------------------------------------------------------------------------------+
    N           Min           Max        Median           Avg        Stddev
x  15          2.99          3.27          3.06         3.076   0.065770163
+  15          2.98          3.36          3.06     3.0973333   0.099388607
No difference proven at 95.0% confidence
$
```

gperciva:
Why the difference?  Well, in scrypt we're "loop-ifying" 3 headers and 3 functions.  Whereas in tarsnap we're "loop-ifying" 54 headers, 67 functions, and...

... hmm, in theory we're telling autoconf that we can loopify 8 decls, but the actual configure script still treats them as separate checks.  *shrug*   I still see something like a 0.8 second reduction in using `AC_CHECK_DECLS_ONCE` in tarsnap, so it's worth it.

PR #333, opened 2022-03-01
Fix strict aliasing violation

gperciva:


gperciva:
Here's a fairly straightforward way of fixing the aliasing violation: pass in `uint32_t *`, for initial comments.

I'm guessing that we'll want to rebase this down to a single commit, but for timing purposes I separated "using memcpy" from "fix the strict aliasing violation".

This PR does not include changing `crypto_scrypt_smix_sse2.c` (yet).

I haven't tried to play any games with `union`.  At least one blog post noted that this method is strictly speaking not allowed in C99, but " it is an extremely common idiom and is well-supported by all major compilers" [1].  I don't know if you want to rely on that.

[1] https://cellperformance.beyond3d.com/articles/2006/06/understanding-strict-aliasing.html

Performance (graph below):
- I tested with `make test N=1`.
- if we use `memcpy` for `blkcpy`, it's a 3% - 4% performance penalty.
- there's absolutely no benefit to replacing blkcpy with memcpy inside salsa20_8 or blockmix_salsa8, so I'm presuming that we want to keep blkcpy for similarity with crypto_scrypt_smix_sse2.c.  (I'm sure that you didn't expect there to be any benefit, but I figured that I might as well measure it just to be certain.)

The plot is a bit busy.  The first 4 columns are clang; the next 4 are gcc.  They both measure master, then the first 3 commits in this PR.  (the last two commits in this PR won't be used; they're just there to demonstrate that there's no point getting rid of blkcpy.)

scrypt-aliasing

gperciva:
Oh, I see.  Quick update; each column is a commit (starting from master), with 3 timing runs.

```
$ more clang.txt 
3.60    3.70    3.91    3.72
3.60    3.71    3.92    3.72
3.61    3.72    3.92    3.75
$ more gcc10.txt 
3.83    3.92    4.05    3.97
3.87    3.92    4.06    3.99
3.86    3.93    4.07    3.99
```

What this means:
- fixing blkxor adds approximately 0.05 to 0.10 seconds
- fixing blkcpy adds approximately 0.15 to 0.20 seconds
- switching blkcpy to memcpy speeds it up by approximately 0.15 seconds

I can make a graph tomorrow, but I suspect that the data is clear enough as text.

gperciva:
Graphs, from slightly different data because I accidentally the previous text files.

clang

gcc

gperciva:
For the record, this does affect tarsnap as well, if somebody:

- is using gcc11 or higher
- compiles with `-O3`
- has a passphrase-protected keyfile

cperciva:
LGTM.  Please tidy up the commit messages, then I'll merge when you say it's ready.  (And also to tarsnap; we should probably roll a 1.0.40.1 for this fix.)

gperciva:
Not quite so fast.  The earlier PR was in a highly "intermediate" step, which was only solving the immediately-reproducible problem.

I've now switched `intergerify()` to use `uint32_t *` (no effect on timings), and used `_mm128i *` for the _sse2.c version (again no effect on timings).

When it comes to memcpy in _sse2.c, clang doesn't care (no effect at all).  For gcc10, there's a -2.3% time change (going from 3.07 to 3.14).  gcc11 takes that down to 3.13 seconds, but I still figure that we won't want to use memcpy for _sse2.c.

lib/crypto/crypto_scrypt_smix.c
```diff
@@ -138,7 +138,7 @@ blockmix_salsa8(const uint32_t * Bin, uint32_t * Bout, uint32_t * X, size_t r)
  * Return the result of parsing B_{2r-1} as a little-endian integer.
  */
 static uint64_t
-integerify(const void * B, size_t r)
+integerify(const uint32_t * B, size_t r)
 {
 	const uint32_t * X = (const void *)((uintptr_t)(B) + (2 * r - 1) * 64);
```
cperciva, line 143: If we're making `B` a `const uint32_t *` then we might as well fix the pointer arithmetic to avoid the now-unnecessary detour into `uintptr_t`.

lib/crypto/crypto_scrypt_smix_sse2.c
```diff
@@ -36,14 +36,14 @@
...
 static void salsa20_8(__m128i[4]);
 static void blockmix_salsa8(const __m128i *, __m128i *, __m128i *, size_t);
-static uint64_t integerify(const void *, size_t);
+static uint64_t integerify(const __m128i *, size_t);
 
 static void
-blkcpy(void * dest, const void * src, size_t len)
+blkcpy(__m128i * dest, const __m128i * src, size_t len)
 {
 	__m128i * D = dest;
```
cperciva: Might as well drop these now?

lib/crypto/crypto_scrypt_smix.c
gperciva: I wondered about that, and I've uploaded it -- but we can only do this in the plain `_smix.c`.  For `_sse2.c`, we have a `__m128i`, so we still need to do the detour into `uintptr_t` (as far as I know).

Having different code in those two functions doesn't look great:
```
/* plain */
const uint32_t * X = B + (2 * r - 1) * 16;

/* sse2 */
const uint32_t * X = (const void *)((uintptr_t)(B) + (2 * r - 1) * 64);
```

But it's your call.

cperciva: We need to drop it in both places.  If the compiler is going to enforce strict aliasing rules, we have to tell it that X is part of B.  I can't remember the intrinsics for extracting a uint32_t from an __m128i though.

gperciva:
Is this what you had in mind?

```
	const __m128i * X = B + (2*r - 1) * 4;
	uint32_t X0, X13;

	X0 = (uint32_t)_mm_cvtsi128_si32(X[0]);
	X13 = (uint32_t)_mm_cvtsi128_si32(X[13]);
	return (((uint64_t)(X13) << 32) + X0);
```

Part of me is thinking "surely not!", but at the moment I can't interpret your comment about extracting uint32_t in any other way.  Also, I can't see another way that's (defnitely?) safe in terms of aliasing.

gperciva:
Whoops, I assumed that testing only `N=1` would find any memory issues.  This problem arose from `test N=2`.

cperciva:
Yes, that or something like it.  I agree that it's ugly, but I think that's the officially correct way of telling the compiler that we're extracting an element from the vector.  Hopefully the assembly code produced is reasonable?

lib/crypto/crypto_scrypt_smix_sse2.c
```diff
@@ -166,9 +166,12 @@ blockmix_salsa8(const __m128i * Bin, __m128i * Bout, __m128i * X, size_t r)
 static uint64_t
 integerify(const __m128i * B, size_t r)
 {
-	const uint32_t * X = (const uint32_t *)(B + (2 * r - 1) * 4);
+	const __m128i * X = B + (2*r - 1) * 4;
+	uint32_t X0, X13;
 
-	return (((uint64_t)(X[13]) << 32) + X[0]);
+	X0 = (uint32_t)_mm_cvtsi128_si32(X[0]);
+	X13 = (uint32_t)_mm_cvtsi128_si32(X[13]);
```
cperciva, line 173: X is an array of 4-element vectors.  So we want the first element of X[0] and the second element of X[3].

gperciva:
Flying Spaghetti Monster.

You're thinking of `_mm_extract_epi32`, which is SSE 4.1.  To get X13, we'll have to do a shuffle, or do `_mm_extract_epi16` twice and reconstruct the uint32.

cperciva:
> You're thinking of `_mm_extract_epi32`, which is SSE 4.1. To get X13, we'll have to do a shuffle, or do `_mm_extract_epi16` twice and reconstruct the uint32.

Shuffle (or shift) followed by _mm_cvtsi128_si32 is probably cleaner than doing two 16-bit extracts.  I agree that it's ugly though. :-(

gperciva:
I'm not certain if you want any git history in this branch.  At the moment I'm guessing not.

There's a tiny performance penalty for gcc10 with sse2, and a small performance penalty for both compilers with the plain C version.

plain

sse2

cperciva:
Ah, the commit message.  Yes, we need more.  How about this:

```
scrypt: Fix strict aliasing

The original scrypt code treated its data as blobs of bytes, accessing them in
whatever manner was convenient from time to time

* as 32-bit words or vectors thereof for the purpose of the Salsa20/8 core
* in machine-word-sized chunks (aka. size_t) for block copy and xor operations
* as 32-bit words for the Integerify function.

This worked fine at the time, but newer compilers apply strict aliasing rules
which allow them to assume that e.g. data accessed as a uint32_t is not the
same as data accessed as a size_t, resulting in miscompilation.

Note that in recent versions of scrypt (after 2015-07-18; versions 1.2.0 and
later) such miscompilation should be detected by the built-in runtime testing.

To avoid aliasing problems, the generic scrypt code now operates on uint32_t
throughout while the SSE2-enabled scrypt code operates on __m128i throughout.

Experimentally, we found that memcpy() speeds up blkcpy() in the plain C
case, but slowed it down in the _sse2.c case (probably because memcpy can make
use of vector instructions internally, but cannot assume that it will always
have a multiple of 16 bytes, as we do).
```

gperciva:
Rebased with that commit message, ready for merge.

cperciva:
Thanks!  You can open a PR to merge this into tarsnap?

PR #334, opened 2022-04-05
Faster plain smix

gperciva:


gperciva:
So I was just about to reply to people on the scrypt mailing list to let them know that the aliasing problem was fixed, but then I noticed that somebody claimed that if we used `unsigned char *` instead of `uint32_t *` then it would use `pxor` and `movdqu` instead of `xorl` and `movl`.

Surely not, I thought.  If the compiler can figure out that `unsigned char *` is aligned, then shouldn't it be able to figure out the same thing about `uint32_t *`?

Apparently not.

Switch `blkxor` to use `unsigned char *` -- which should still be valid with strict aliasing -- gives us:

* clang: 3% improvement
* gcc10 (and 11): 7.5% improvement.

scrypt

On the minus side, this complicates the code in `blockmix_salsa8`, in terms of casting the `uint32_t *` variables to `unsigned char *` for the `blkxor` calls.

I'm not certain if this is worth it or not, since it applies to the non-sse2 code.  I also haven't tested it on ARM (yet?).

cperciva:
I'm inclined to not bother.  Now that we autodetect SSE2 support (which the original code didn't do) approximately 0% of x86 users should be running the non-SSE2 code, and we would need to benchmark a range of different non-x86 systems (and compilers) to get a sense of whether a change would help there.

PR #337, opened 2022-04-26
style: space in "switch (" and "while ("

gperciva:


gperciva:
I wouldn't normally make a PR just for something this small & trivial, but I'd like to keep the libcperciva import PRs in sync, and I'm doing them for spiped / tarsnap / kivaloo.

PR #339, opened 2022-05-25
build: improve autoconf compatibility

gperciva:


gperciva:
This has absolutely no effect with autoconf 2.69, but will eliminate an amazingly long & convoluted warning on autoconf 2.71+.

PR #340, opened 2022-05-31
Libcperciva import and clang nitpicks

gperciva:


gperciva:
First two commits are imports from libcperciva; the next two commits are the scrypt equivalents.

PR #343, opened 2022-08-15
Build: move more configure.ac into m4/

gperciva:


gperciva:
This allow us to share more build-system code between scrypt and tarsnap.  No timing difference with autoconf 2.69 or 2.71.

PR #348, opened 2022-09-15
Remove underscores in #ifndef #define #endif (scrypt)

gperciva:


gperciva:
This does not include modifying the github actions to remove `-Wno-reserved-id-macro` because scrypt is more complicated in that regard, due to `memlimit.c` and autoconf's `config.h`.

PR #349, opened 2022-09-20
memlimit.c: remove leading underscores; handle non-POSIX in build system

gperciva:


gperciva:
Same rational as the apisupport NONPOSIX SETGROUPS check in https://github.com/Tarsnap/libcperciva/pull/445.

I considered adding the whole apisupport infrastructure, but in the end I decided not to.  At the moment, the only relevant apisupport check would be "should we remove `_POSIX_C__SOURCE` and `_XOPEN_SOURCE from `memlimit.c`?", and previously we've been unconditionally doing so.  I don't see the benefit in adding a bunch of additional build complexity just to avoid undefining those values if the compiler isn't being nitpicky.

If we add apisupport for some other reason, then we could reconsider this (although my first instinct would still be to say that there's no harm in unconditionally undefining these).  Of course, if there's any platform which requires those terms (which would be a "wat." situation), then I'd be happy to add whatever amount of build system complexity that was required.

PR #355, opened 2022-12-19
scryptenc: add _print_error() to handle error strings

gperciva:


gperciva:
I think this `switch()` should at least be a `static void` in `main.c`, because that would let us (easily) print errors at the appropriate spot, namely:

```
if ((rc = scryptenc_foo()) != SCRYPT_OK) {
        scryptenc_print_error(rc, infilename, outfilename));
        goto err2;
}
```

instead of keeping `rc` around and printing its errors at the very end.

I originally had this as a `static void` function in `main.c`, but the more I looked at that, the more I thought it would be better in `scryptenc.c`, because the error codes are defined in `scryptenc.h`.

We've previously mused about having a `scryptenc_errno.h` https://github.com/Tarsnap/scrypt/pull/214#discussion_r341890487.  I don't think it's necessary, but I'm open to shoving this in there (perhaps as an `inline void`, a la `millisleep()` in spiped?).

gperciva:
> I'm not super enthusiastic about scryptenc.c reporting errors by calling warnp directly

... then it's a good thing that this PR doesn't do that?  `rc = scrypt...` is still in `main.c`.  The `SCRYPT_foo` defines are still in `scryptenc.h`.  Nothing stops an application from treating them purely as error codes.

> How about a scryptenc_strerror function which returns an error string? (Created by asprintf, owned by scryptenc and freed via atexit, so that you can warnp("%s", scryptenc_strerror(...)) without worrying about memory leaks.)

If we're dynamically allocating the string via `asprintf()`, then we could theoretically fail to report the correct error message and presumably would output a generic "out of memory required to print the real error" string.  (Granted, I can't think of any plausible scenario that would involve not being able to allocate a dozen bytes.  And if the system were that overloaded, then `fprintf` within `warnp.c` could also fail with `ENOMEM`.  But this still seems like it would be a puzzling design choice!)

Another possibility would be to pre-allocate a string long enough to handle any possible `infilename` or `outfilename`, but that runs into whether platforms actually include `PATH_MAX` in `<limits.h>` or not.

gperciva:
> The SCRYPT_foo defines are still in scryptenc.h. Nothing stops an application from treating them purely as error codes.

To emphasize that point, I could rewrite this somewhat:

- continue to keep the `SCRYPT_foo` defines in `scryptenc.h`.  (in other words, this PR would not modify `scryptenc.h` or `scryptenc.c` at all.)
- add a new `scryptenc_errno_print.h` which contains the relevant function.

That way, it's absolutely clear that `scryptenc.c` is not calling `warnp` directly, and in fact, could be used in a project which does not contain `warnp.c` at all.

PR #357, opened 2022-12-30
scryptenc_errno.h

gperciva:


gperciva:
This is an alternate version of #355; rather than adding `scryptenc_print_error()` to `scryptenc.c`, it introduces a new `scryptenc_errno.h`.

I don't think that you'll like it, but I thought it would be useful to have the code here if you wanted to take a quick look and compare that with #355.

(I don't hate this version as much as I expected to.  I kind-of think that if we're going to have `scryptenc_errno.h` then we might as well add `scryptnec_errno.c` because that `switch` looks too big to justify as a `static inline void`.  That said, I think that #355 is a better solution than this.)

PR #358, opened 2023-01-16
Add scryptenc_print_error() as a separate file(s)

gperciva:


gperciva:
This implements my musing in https://github.com/Tarsnap/scrypt/pull/new/scryptenc-print-error-separate.  It clarifies that adding a function to print scryptenc error codes does not modify `scryptenc.h` or `scryptenc.c` at all.

If you like this sketch, you might prefer to have all the code in the `.h` file (a la `millisleep.h`); I'm happy to make that change.

cperciva:
Ok, I can live with this.  I still wonder if a future user of this code would want to get error strings which they can output in some other way, but that's a hypothetical and it's not worth spending time on until/unless someone actually comes forward with a desire to do that.

gperciva:
If we weren't specifying the `infilename` or `outfilename`, I'd 100% make a `const char * scryptenc_strerror()`.  But as it is, those strings would need to include a `%s`, and the error-printing would need to special-case handling `SCRYPT_WRFILE` and `SCRYPT_ERDFILE`, and that just gets really messy.

As you say, we can switch to another way of doing this later, if there's a well-specified desire for it.

PR #359, opened 2023-01-25
Refactor main.c with scrypt_mode_{info,enc_dec}

gperciva:


gperciva:
I'm not a fan of the existing clean-up in `main.c`.  Here's two solutions to it:
- #359 adds `scrypt_mode_info()` and `scrypt_mode_enc_dec()` as static functions in `main.c`.
- #360 adds those same functions into a new `scrypt_mode.c` file.

In both cases, look at how much cleaner `main()` is!

(I have no particular thoughts about which option is better.)

main.c
```diff
@@ -280,91 +383,17 @@ main(int argc, char *argv[])
 		/* User selected 'info' mode. */
 		if (scrypt_mode_info(infilename))
 			goto err0;
-		rc = SCRYPT_OK;
-		goto done;
-	}
-
-	/* If the input isn't stdin, open the file. */
```
cperciva, line 236: Why not keep this here?  I don't see any benefit to duplicating this block between the two scrypt_mode functions.

cperciva:
I prefer this version.  Not sure why you duplicated the file-opening code though...?

gperciva:
I've sketched a variant where we keep `fopen(infile)` in `main()`.  See
https://github.com/Tarsnap/scrypt/compare/refactor-main-fopen?expand=1

I'm not wild about it:

- passing both `infile` and `infilename` to functions seems silly, i.e.:
    ```
    /**
     * scrypt_mode_info(infile, infilename):
     * Print scrypt parameters used for the specified ${infile} and ${infilename}.
     * If ${infile} is stdin, ${infilename} must be NULL.  Close ${infile} when
     * finished.
     */
    ```
    We need `infilename` in that function for `scryptenc_print_error()`.

- as mentioned in the function comment, the `scrypt_mode_*` functions do `fclose(infile)`, which is a different function from where the `fopen()` happens.  That's not a huge problem, but it seems a bit unfortunate.

- Other than `FILE * infile`, all of the variables in `main()` are directly related to parsing the command line.

I also made a sketch where we keep the `fclose(infile)` in `main()`.  That slightly simplifies the function comment -- no mention of closing `${infile}` -- but then there's the oddness of `scrypt_mode_enc_dec` opening and fclosing `outfile` but not `infile`.

All told, only giving `infilename` to the functions would reduce the amount of state being passed around, and therefore would be more readable.  I figured that the duplicate `fopen()` code in local functions was worth it.

PR #360, opened 2023-01-25
Add new scrypt_mode.{c,h} for scrypt_mode_{info,enc_dec}

gperciva:


gperciva:
I'm not a fan of the existing clean-up in `main.c`.  Here's two solutions to it:
- #359 adds `scrypt_mode_info()` and `scrypt_mode_enc_dec()` as static functions in `main.c`.
- #360 adds those same functions into a new `scrypt_mode.c` file.

In both cases, look at how much cleaner `main()` is!

(I have no particular thoughts about which option is better.)

PR #373, opened 2023-08-07
Libcperciva import and trivial sh style

gperciva:


gperciva:
In addition to the libcperciva import, this:
- adds double-quotes to a variable in `get-version.sh`
- adds curly braces {} to variables in `release-tools/create-sign-tarball.sh` and `mktarball.sh`.

PR #374, opened 2023-08-29
Libcperciva import

gperciva:


gperciva:
This does not include `optional_mutex` because scrypt doesn't need it.

PR #378, opened 2023-10-30
build: add checks for openssl dir via brew on macOS

gperciva:


gperciva:
Imported from https://github.com/Tarsnap/tarsnap/pull/583.

gperciva:
Hmph.  The current image for macos for Github actions is documented as using openssl 1.1.1, but based on the compiler errors, it's clearly 3.x.

[1] https://github.com/actions/runner-images/releases/tag/macOS-11%2F20230921.4

PR #391, opened 2024-08-10
scryptenc_print_error.c: add a "default" for safety

gperciva:


gperciva:
We should never need this, but there's no harm in adding it.

gperciva:
FWIW, this was inspired by tarsnap's `lib/keyfile/keyfile.c`, which includes a `default:` whenever checking the results of `scrypt{enc,dec}_buf()`.

https://github.com/Tarsnap/tarsnap/blob/master/lib/keyfile/keyfile.c#L199

PR #394, opened 2024-12-21
libcperciva import, fix Makefile.am, and fix test

gperciva:


gperciva:
There was one report of this test failing because it successfully decrypted the 10-second file in 1 second [*].  This was on a Debian build server, so I suspect that the system was heavily loaded while it created the 10-second file, so `scrypt enc` underestimated the parameters.

[*] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1086624

PR #395, opened 2025-01-29
Improve comments, fix dates, add header

gperciva:


gperciva:
I originally wrote this PR as 4 separate commits (3 imports from libcperciva, 1 new).  But two of those commits were only one line, and they were all trivial, so I rebased them all together.

PR #397, opened 2025-01-30
build: fix name of crypto_aes library

gperciva:


gperciva:
This includes #396.

We made a mistake in #386 -- this library only includes libcperciva code, not scrypt, so it should be `libcperciva_*`.

gperciva:
I think this one is ready to go.  (The "sorting library order" PR still needs work.)

This does:
- change name
- remove `crypto_aes.h` from `_SOURCES`
- add dependency on `cpusupport-config.h`

This library's `CFLAGS` does not include `cpusupport-config.h` compiler flags (unlike all the other libcperciva libraries for scrypt), but that checks out -- we don't need anything like `-maes` to compile this file.  We *do* need defines such as `CPUSUPPORT_X86_AESNI` to compile it correctly, but those come from the `#include "cpusupport.h"`.

PR #398, opened 2025-01-30
Makefile.am: sort lists

gperciva:


gperciva:
This depends on #396 and #397.

gperciva:
When we added `libscrypt_crypto_aes` in #386, it worked almost by accident.  That is: since scrypt uses libtool, it's a lot more forgiving of the order of libraries.  Also, clang's linker is forgiving.

If we try to apply this method of `lib...crypto_aes` to tarsnap with gcc's linker, it fails unless we add libtarsnap.a to _LDADD twice.  Once for the general "here's the symbols we need", and once at the end for "here's the `cpusupport_x86_aesni_detect_1` etc symbols".

I only got that working 10 minutes ago, so I'm still reconsidering how best to approach it in scrypt.  (But the PRs you already merged are still good.)

PR #399, opened 2025-01-31
Makefile.am: add libcperciva_cpusupport_detect.la

gperciva:


gperciva:
I'm reasonably confident that the diff is correct.

I'm not confident that the commit message does an adequate job of explaining it; particularly since this is the scrypt repository, not the tarsnap one.

libcpusupport_crypto_aes.la needs to resolve symbols such as cpusupport_x86_aesni_detect_1.  In scrypt, that was previously provided by cpusupport_x86_aesni.o, which was listed explicitly on the command-line (via Makefile.am's ${crypto_scrypt_files}).

However, in the tarsnap repository, cpusupport_x86_aesni.o was linked into libtarsnap.a.  When we tried to add libcperciva_crypto_aes.a, that created a circular dependency:
- libtarsnap.a needed crypto_aes symbols (which were in libcperciva_crypto_aes)
- libcperciva_crypto_aes needed cpusupport_x86_aesni_detect_1 (which were in libtarsnap.a)

To avoid the circular dependency, we're moving the cpusupport detection functionality into its own library.

gperciva:
Wait, I should check that this actually *does* fix the problem for tarsnap.

gperciva:
Yep, it fixes the tarsnap problem.

Makefile.am
```diff
@@ -101,15 +91,20 @@ AM_CPPFLAGS=								\
 	-D_XOPEN_SOURCE=700						\
 	${CFLAGS_POSIX}
 
+# Each "paragraph" denotes a rank of dependencies.  Libraries without
```
gperciva, line 94: This is wordier than I'd like, but I decided to err on the side of clarity.

gperciva:
Nope, I forgot to check libscrypt-kdf.  (and it's broken)

I'm beginning to have some mild dislike of autotools.

gperciva:
Fortunately it was an easy fix, and everything appears to be working now.

gperciva:
This was merged as part of #398.

PR #403, opened 2025-12-22
actions: update to macos-14 (arm64)

gperciva:


gperciva:
scrypt and tarsnap don't have hard-coded (in the github Actions file) paths for openssl, so we don't need to change those directories -- autoconf already finds the directories.

PR #404, opened 2026-04-17
libcperciva import

gperciva:


gperciva:
No build stuff here because it's autoconf, not metabuild.sh.

A rather small PR, but I figured it was worth syncing up, and it's been over a year.

PR #405, opened 2026-05-20
Validate reference scrypt parameters

achimala:
The reference `crypto_scrypt()` implementation was less strict than the primary implementation for invalid KDF parameters:

- `r == 0` and `p == 0` reached later allocation/math checks and failed with `ENOMEM` in my local harness rather than `EINVAL`.
- `N == 1` was accepted even though the public API documents `N` as a power of two greater than 1 and the primary implementation rejects `N < 2`.

This mirrors the existing guards in `lib-platform/crypto/crypto_scrypt.c`, updates the reference comments/docs to state `0 < r * p < 2^30`, and adds a small regression binary for the reference implementation.

Verification on macOS/Apple Silicon from a no-space build path (`/tmp/scrypt-verify`):

```sh
PATH="/opt/homebrew/opt/libtool/libexec/gnubin:$PATH" autoreconf -i
CPPFLAGS="-I/opt/homebrew/opt/openssl@3/include" \
  LDFLAGS="-L/opt/homebrew/opt/openssl@3/lib" \
  ./configure --enable-libscrypt-kdf
make -j4
make test SMALLMEM=1
```

Result: the new `tests/verify-strings/test_scrypt_ref_params` binary passed, and the existing test scenarios passed; `05-system-scrypt-encrypt-decrypt` skipped because no suitable system `scrypt` was installed.

I used Codex assistance to find, patch, and test this change.

PR #407, opened 2026-05-30
Reject N less than two in reference scrypt

achimala:
## Summary
- reject `N < 2` in the reference `crypto_scrypt()` parameter validation
- update the internal comments to state that `N` must be a power of two greater than 1

## Tests
- `git diff --check`

Split out from #405.


gperciva:
Thanks, code looks great!

One nitpick in the git commit message: please wrap the lines less than 80 chars.  I tend to wrap at 76 or 78 chars.

achimala:
Thanks, updated. I amended the commit message to wrap before 80 chars.

PR #406, opened 2026-05-30
Reject zero r or p in reference scrypt

achimala:
## Summary
- reject `r == 0` or `p == 0` in `crypto_scrypt()` with `EINVAL`
- document the lower bound as `0 < r * p < 2^30` in `FORMAT`, `README.md`, `scrypt-kdf.h`, and the reference implementation comment

## Notes
This follows up on 686a628f995d86685d87045c80fba0ffce318c71 as requested in #405.

## Tests
- `git diff --check`

Split out from #405.

PR #408, opened 2026-05-30
Add reference scrypt parameter validation test

achimala:
## Summary
- add a dedicated invalid-parameter test for the reference `crypto_scrypt()` implementation
- cover zero `r`, zero `p`, and `N == 1`
- place the test under `tests/invalid-params/` rather than `tests/verify-strings/`

## Dependency
This is intentionally opened as a draft because the test is expected to pass after the validation fixes in #406 and #407 are applied.

## Tests
- `git diff --check`

Split out from #405.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,5 +1,6 @@
 bin_PROGRAMS=	scrypt
 noinst_PROGRAMS=							\
+	tests/invalid-params/test_scrypt_ref_params			\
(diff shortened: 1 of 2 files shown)


gperciva:
Thanks for this submission.  I ended up implementing it in a rather different way, inhttps://github.com/Tarsnap/scrypt/pull/419.

By the way, your bug bounties so far come to $40.  (two $10, one $15, one $5)

PR #410, opened 2026-06-30
AGENTS.md: add

gperciva:


gperciva:
This is like https://github.com/Tarsnap/tarsnap/pull/693, except that I removed any mention of upstream code, because we don't have that in scrypt.

PR #411, opened 2026-07-10
bug: Encrypt-then-MAC violation in scryptdec_buf — outlen set before HMAC verification (#702)

lushan888:
### Vulnerability
`scryptdec_buf()` decrypts ciphertext into `outbuf` and sets `*outlen` **before** verifying the HMAC signature. This violates the Encrypt-then-MAC principle. An attacker who tampers with encrypted data can cause unauthenticated plaintext to be exposed to the caller, even when the function ultimately returns an error.

### Fix
1. Move `*outlen = inbuflen - 128` to **after** HMAC verification succeeds
2. Changed HMAC failure to `return (SCRYPT_EINVAL)` with `insecure_memzero` instead of `goto err1`

### Verification
```c
/* Before (vulnerable): */
crypto_aesctr_stream(AES, &inbuf[96], outbuf, inbuflen - 128);
*outlen = inbuflen - 128;  // ← Caller can read unauthenticated data!
/* Verify signature */
if (crypto_verify_bytes(hbuf, &inbuf[inbuflen - 32], 32)) {
    rc = SCRYPT_EINVAL;
    goto err1;  // ← outbuf already set, *outlen already written!
}

/* After (fixed): */
crypto_aesctr_stream(AES, &inbuf[96], outbuf, inbuflen - 128);
/* Verify signature before setting *outlen (encrypt-then-MAC) */
if (crypto_verify_bytes(hbuf, &inbuf[inbuflen - 32], 32)) {
    insecure_memzero(dk, 64);
    return (SCRYPT_EINVAL);  // ← outbuf still set, but *outlen never written!
}
*outlen = inbuflen - 128;  // ← Only set after authentication
```

**Wallet (Base):** 0xaa9e4971e0973065513b18dEF9eC06Eb1F39D7A2

### DIFF
--- a/lib/scryptenc/scryptenc.c
+++ b/lib/scryptenc/scryptenc.c
@@ -81,11 +81,13 @@ display_params(int logN, uint32_t r, uint32_t p, size_t memlimit,
 	fprintf(stderr, "Parameters used: N = %" PRIu64 "; r = %" PRIu32
 	    "; p = %" PRIu32 ";\n", N, r, p);
 
-	/* Memory */
+	/* Memory - guard against humansize() returning NULL on ENOMEM */
 	fprintf(stderr, "    Decrypting this file requires at least"
-	    " %s of memory", human_mem_minimum);
+	    " %s of memory",
+	    human_mem_minimum ? human_mem_minimum : "(unknown)");
 	if (memlimit > 0)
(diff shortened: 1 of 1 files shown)


cperciva:
Better fixes: #412 #413

PR #412, opened 2026-07-10
scryptenc: check for NULL after humansize() use

gperciva:


lib/scryptenc/scryptenc.c
```diff
@@ -96,6 +100,16 @@ display_params(int logN, uint32_t r, uint32_t p, size_t memlimit,
 	/* Clean up */
 	free(human_memlimit);
 	free(human_mem_minimum);
+
+	/* Success! */
+	return (0);
```
cperciva: `SCRYPT_OK`?

lib/scryptenc/scryptenc.c
```diff
@@ -96,6 +100,16 @@ display_params(int logN, uint32_t r, uint32_t p, size_t memlimit,
...
+
+err1:
+	free(human_memlimit);
+	free(human_mem_minimum);
+
+	/* Failure! */
+	return (-1);
```
cperciva: `SCRYPT_ENOMEM`?

cperciva:
If we're going to pass the return code from `display_params` upstream, it needs to be a `SCRYPT_*` value.

And I'm thinking $10 for this one; "crash instead of exiting with an error" is arguably cosmetic, and in this particular case I can't imagine how it could be triggered by accident... in fact I'm not even sure how it could be triggered *deliberately* without a nerfed malloc.

lib/scryptenc/scryptenc.c
gperciva: Whoops, I had that in, then took it out, then mis-added it back.

gperciva:
Added REBASE commit

cperciva:
Fix looks good.  Please tidy up the history and set the bounty to $10.

PR #413, opened 2026-07-10
scryptdec_buf: Verify HMAC first

cperciva:
Callers of this API should treat output parameters as undefined unless the function returns success; but it nonetheless makes sense to check the HMAC first and error out early.

Reported by:	ChronoNova5678
Bug Bounty:	$10 (harmless)


cperciva:
Fixes https://github.com/Tarsnap/tarsnap/issues/702

gperciva:
Code looks good.

In the comment, in the humansize NULL issue, the reporter asked to be credited as "ChronoNova" (with no numbers); but there was no such request for this particular issue.  My initial instinct would be to credit both bounty reports to `ChronoNova`.

cperciva:
Thanks, reporter name fixed.

PR #418, opened 2026-07-21
scryptenc: fix display_params() calls

gperciva:


gperciva:
This fixes my mistakes in #412.  I even remember wondering "am I sure that's enough parentheses?", but I failed to double-check it.

PR #419, opened 2026-07-21
Add check_params tests

gperciva:


gperciva:
This uses the same source file, but links it to either the normal scrypt stuff, or a reference-files-only library (with no intrinsics).

cperciva:
Can you split adding the C file off into a separate commit, and/or add text to the commit messages to clarify the "same source file is used by both tests" situation?  I think I understand it now but it took me long enough that I feel it should be clearer.

gperciva:
I've updated the commit message of the second commit.  I'd rather not add the `.c` file by itself, since that's not a self-contained commit.

As it stands now, in theory the first commit could be merged on its own without leaving the tree in an un-sensible state.

gperciva:
Thanks!

BTW, when a PR has multiple commits, you normally merge with a merge commit, rather than a linear-history rebase.

I'm not fussed either way, but I figured I'd mention it in case it wasn't an intentional change.

cperciva:
Yeah, I could have gone either way on this one.  FreeBSD uses a rebase-only model which has made me lean more towards using rebase for this sort of "two commits of which the first can stand entirely independently".  If you had turned it into three commits with the first one being "add a C file which will be used later" then I would definitely have merged in order to avoid having a commit which didn't stand by itself.

PR #420, opened 2026-07-21
Misc improvements to github actions

gperciva:


.github/workflows/compile.yml
```diff
@@ -111,3 +116,10 @@ jobs:
       run: make CFLAGS="${{ env.CFLAGS }}"
     - name: Test clang binaries
       run: make test VERBOSE=1
+    - name: Check for untracked files
```
gperciva, line 119: When I initially wrote that file, I decided that I'd "save computational resources" by skipping the "untracked files" check on macos.  But that was ridiculous; `git status` is trivial compared to spinning up a VM, compiling, running tests, etc.  Might as well check for unexpected files on both OSes.

PR #428, opened 2026-09-06
main: refuse to write the output into the input file

### LINKED ISSUE #427: [bug bounty] "scrypt enc f f" silently destroys f: the output file is truncated before the input is read

### Summary

`scrypt enc secret.txt secret.txt` destroys `secret.txt` and exits 0.  The file
is replaced by a 128-byte scrypt file whose plaintext is empty; the original
contents are gone, and nothing is printed to say so.

### What happens

`scrypt_mode_enc_dec()` in `main.c` opens the input, reads the passphrase, and
only then opens the output:

```c
	/* If the input isn't stdin, open the file. */
	if (infilename != NULL) {
		if ((infile = fopen(infilename, "rb")) == NULL) {
...
	/* If we have an output filename, open it. */
	if (outfilename != NULL) {
		if ((outfile = fopen(outfilename, "wb")) == NULL) {
```

When the two names refer to the same file, `fopen(outfilename, "wb")`
truncates it to zero length.  Nothing has been read from `infile` yet — the
first `fread()` in `scryptenc_file()` happens afterwards — so that `fread()`
returns 0 immediately:

```c
	do {
		if ((readlen = fread(buf, 1, ENCBLOCK, infile)) == 0)
			break;
```

`ferror(infile)` is false, so this is indistinguishable from a legitimately
empty input.  `scryptenc_file()` writes the 96-byte header and the 32-byte
HMAC, returns `SCRYPT_OK`, and `scrypt` exits 0.

### Reproducer

```sh
$ echo "important data" > secret.txt
$ echo "hunter2" | scrypt enc --passphrase dev:stdin-once secret.txt secret.txt
$ echo $?
0
$ ls -l secret.txt
-rw-r--r--  1 user  user  128 ... secret.txt
$ echo "hunter2" | scrypt dec --passphrase dev:stdin-once secret.txt
$ echo $?
0
```

The final decryption succeeds and prints nothing: the file now correctly
contains an encryption of zero bytes.  The original 15 bytes are unrecoverable.

`scrypt dec f f` loses the ciphertext the same way — the header is read into
the cookie by `scryptdec_file_prep()`, then the file is truncated, then
`scryptdec_file_copy()` fails on the empty remainder — leaving `f` as a
zero-length file.

Both are also reachable with the input on stdin:

```sh
echo "hunter2" | scrypt enc --passphrase dev:stdin-once - f < f
```

### Why I think this is worth fixing rather than "don't do that"

`scrypt` opens both files itself; the shell is not involved, so this is not the
`sort f > f` situation where the shell truncates before the program runs.  A
user asking to encrypt a file in place is a natural thing to try, `scrypt(1)`
does not say it is unsupported, and the failure is silent and total —
`cp`, `mv` and `install` all diagnose the same mistake instead.

### Fix

PR to follow: compare the open input file and the output path with `fstat()`
and `stat()` before opening the output, and exit with an error if they are the
same regular file.  Devices are excluded, since opening `/dev/null` for writing
discards nothing.  The check runs before the passphrase is read, so the user is
not prompted for a passphrase that cannot be used.

It adds `tests/12-same-file.sh` and a note in `scrypt.1`.

### How this was found

Source review of `main.c` and `lib/scryptenc/scryptenc.c` at `a71ae82`,
following the ordering of the `fopen()` calls against the first `fread()`.

woahwhattheheck:

Fixes #427.

`scrypt enc secret.txt secret.txt` opens `secret.txt` for reading, reads a
passphrase, then opens the same path with `fopen(..., "wb")` — which truncates
it.  Nothing has been read from the input yet, so the first `fread()` in
`scryptenc_file()` returns 0, `ferror()` is false, and the empty read is
indistinguishable from a legitimately empty file.  `scryptenc_file()` writes the
header and HMAC, returns `SCRYPT_OK`, and `scrypt` exits 0.  The user's file is
replaced by a 128-byte encryption of nothing, silently.

`scrypt dec f f` loses the ciphertext the same way, and both are reachable with
the input on stdin (`scrypt enc - f < f`).

### The change

`same_file()` compares the already-open input against the output path with
`fstat()`/`stat()` and reports whether they are the same regular file:

* It uses `fstat(fileno(infile))` rather than `stat(infilename)` so that the
  stdin cases are covered too, and so that symlinks, hard links and different
  spellings of the same path are all handled.
* Only regular files count — opening a character device such as `/dev/null`
  for writing discards nothing, so `scrypt enc /dev/null /dev/null` keeps
  working.
* If either `stat` fails the answer is "different", so the normal case where
  the output file does not exist yet costs one failed `stat()` and nothing
  else.

The check sits immediately after the input is opened and **before** the
passphrase is read, so the user is not asked to type a passphrase for a command
that cannot succeed.  It jumps to the existing `err1` label, which closes the
input; no output file is created.

`fileno()` is already used in `libcperciva/util/readpass.c` and `<sys/types.h>`
in `lib-platform/util/memlimit.c`, so this adds no new portability burden.

### Also

* `tests/12-same-file.sh` covers `enc` and `dec` with the same file, asserts
  exit status 1, the error message, and that the input is byte-for-byte
  unchanged; plus a check that encrypting to a *different* file still works, so
  a false positive in `same_file()` would be caught.
* `scrypt.1` documents the restriction.

### Note on the test number

The test is numbered `12` rather than `11` so that it does not collide with
`tests/11-info.sh` from #426.  The ordering works whichever of the two lands
first; the only overlap is the adjacent line each adds to `EXTRA_DIST` in
`Makefile.am`.  Happy to renumber if you would rather not take #426.


tokenjunkielabs:
Head `14db0e0600340f8254fe602fd07e25a8fe3f59ad` adds a deterministic regression for the TOCTOU seam already handled by `open_output()`: the test starts with a distinct output, gates passphrase reading on a FIFO so the early pathname check has already passed, swaps the output path to a hard link of the input, then resumes and requires exit 1, the same-file diagnostic, and byte-for-byte input preservation. The added shell block parses under sh/bash/dash, and its FIFO/path-swap synchronization was exercised independently. Upstream Compile & test run 35535619823 registered for this exact head but is `action_required` with no executed jobs, so this is not a hosted-green claim.

tokenjunkielabs:
Maintainer review / bounty-classification request for #428 (fixes first-reporter report #427). The reported behavior is destructive: `scrypt enc f f` can truncate the input before it is read, destroying the source data; the current PR carries the same-file/alias guard and race-safe follow-up. Please review and classify #427 under the published Tarsnap bug-bounty schedule, including payment for first reporter `woahwhattheheck` if accepted. The schedule lists $500 for data corruption/loss; sponsor classification controls and I am not treating that amount as awarded.

PR #432, opened 2026-09-06
readpass_file: reject a passphrase file with a stray carriage return

### LINKED ISSUE #431: [bug bounty] readpass_file() silently truncates the passphrase at a carriage return

### Summary

`readpass_file()` silently truncates the passphrase at the first carriage
return, so `--passphrase file:FILENAME` can encrypt or decrypt with a
passphrase which is not the one the file contains.

Its documented contract in `libcperciva/util/readpass.h` says such a file
should be rejected:

> Print an error and fail if the file is 2048 characters or more, **or if it
> contains any newline \n or \r\n characters other than at the end of the
> file**.  Do not include the \n or \r\n characters in the passphrase.

### What happens

`libcperciva/util/readpass_file.c`:

```c
	/* Get a line from the file. */
	if ((fgets(passbuf, MAXPASSLEN, f)) == NULL) {
...
	/* Bail if there's the line is too long, or if there's a second line. */
	if (fgetc(f) != EOF) {
...
	/* Truncate at any newline character. */
	passbuf[strcspn(passbuf, "\r\n")] = '\0';
```

`fgets()` stops after a `\n`, so the "more than 1 line" check catches a stray
`\n`: any `\n` we hold is necessarily the last character.  A `\r` is different
— `fgets()` reads straight past it — so a `\r` anywhere in the line reaches
`strcspn()`, which cuts the passphrase off at that point without a word.

### Reproducer

The passphrase file below begins with the correct passphrase, so the
truncation is invisible: the decryption succeeds using a passphrase which is
not what the file holds.

```sh
$ printf 'hunter2\rextra\n' > pw.txt
$ scrypt dec --passphrase file:pw.txt reference.enc out.txt
$ echo $?
0
```

`reference.enc` here is `tests/verify-strings/test_scrypt_good.enc`, which was
encrypted with `hunter2`.  The file says the passphrase is `hunter2\rextra`;
scrypt used `hunter2`.

The same applies when encrypting, and because the truncation is applied
identically both ways it round-trips, so nothing reveals it until the
passphrase is entered by some other route — typed at the prompt, or via
`--passphrase env:VAR` — and then does not work.

A passphrase file whose only newline is a `\n` or a trailing `\r\n` is
unaffected, so the ordinary Unix and Windows cases are both fine today.

### Fix

PR to follow: strip a trailing `\n` or `\r\n` explicitly, then reject the file
if any `\r` or `\n` remains, which is what the header already promises.
Cutting the passphrase short and keeping the character as passphrase data both
change the passphrase an existing file yields, so neither is safe to pick
silently.

It extends `tests/08-passphrase-file.sh` rather than adding a new test file.

### How this was found

Source review of `libcperciva/util/readpass_file.c` at `a71ae82` against the
contract in `libcperciva/util/readpass.h`.

woahwhattheheck:

Fixes #431.

`readpass_file()` ends with

```c
	/* Truncate at any newline character. */
	passbuf[strcspn(passbuf, "\r\n")] = '\0';
```

`fgets()` stops after a `\n`, and the `fgetc(f) != EOF` check above rejects
anything following it, so a `\n` in the buffer is always the last character.
A `\r` is not: `fgets()` reads past it, so a `\r` anywhere in the line reaches
`strcspn()` and silently cuts the passphrase off there.

`--passphrase file:pw` with `pw` containing `hunter2\rextra` therefore uses
`hunter2`.  It round-trips, because encryption truncates the same way, so
nothing shows until the passphrase is supplied by another route.  The header
in `libcperciva/util/readpass.h` already says a file like that should be
rejected:

> Print an error and fail if the file is 2048 characters or more, or if it
> contains any newline \n or \r\n characters other than at the end of the
> file.

### The change

Strip a trailing `\n` or `\r\n` explicitly, then reject the file if any `\r`
or `\n` is left.  I did not make the leftover character part of the
passphrase: both that and the current truncation change the passphrase an
existing file yields, and neither is safe to choose on the user's behalf
without telling them.  No documentation change — this makes the code match
what `readpass.h` already describes.

Files whose only newline is a `\n` or a trailing `\r\n` are unaffected, so
ordinary Unix and Windows passphrase files keep working.

### Test

`tests/08-passphrase-file.sh` gains four checks rather than a new test file:

* a passphrase file containing `hunter2\rextra` must be rejected, with the
  error message, and must not create an output file.  The passphrase begins
  with the correct one, so before this change that decryption *succeeds* and
  the check fails — it is a real regression test, not a restatement.
* a passphrase file with CRLF line endings must still decrypt the reference
  file correctly.

### Note

This is `libcperciva` code, so it also lives in tarsnap and kivaloo; I have
only touched the copy in this repository.

### DIFF
--- a/libcperciva/util/readpass_file.c
+++ b/libcperciva/util/readpass_file.c
@@ -22,6 +22,7 @@ readpass_file(char ** passwd, const char * filename)
 {
 	FILE * f;
 	char passbuf[MAXPASSLEN];
+	size_t len;
 
 	/* Open the file. */
 	if ((f = fopen(filename, "r")) == NULL) {
@@ -52,8 +53,28 @@ readpass_file(char ** passwd, const char * filename)
 		goto err1;
 	}
 
-	/* Truncate at any newline character. */
-	passbuf[strcspn(passbuf, "\r\n")] = '\0';
+	/*
+	 * Strip a trailing "\n" or "\r\n".  fgets() stops after a "\n", so
+	 * if the file contains one then it is the last character we read.
+	 */
+	len = strlen(passbuf);
+	if ((len > 0) && (passbuf[len - 1] == '\n')) {
+		passbuf[--len] = '\0';
+		if ((len > 0) && (passbuf[len - 1] == '\r'))
+			passbuf[--len] = '\0';
+	}
+
+	/*
+	 * Any "\r" or "\n" left is not a line ending we strip.  We could cut
+	 * the passphrase short there, or keep the character as part of the
+	 * passphrase, but both change the passphrase which an existing file
+	 * yields, and neither is safe to pick silently.  Refuse the file.
+	 */
+	if (strcspn(passbuf, "\r\n") != len) {
+		warn0("passphrase in %s contains a carriage return or newline",
+		    filename);
+		goto err1;
+	}
 
 	/* Copy the password out. */
 	if ((*passwd = strdup(passbuf)) == NULL) {
--- a/tests/08-passphrase-file.sh
+++ b/tests/08-passphrase-file.sh
@@ -10,6 +10,7 @@ decrypted_reference_file="${s_basename}-attempt_reference.txt"
 decrypted_badpass_file="${s_basename}-decrypt-badpass.txt"
 decrypted_badpass_log="${s_basename}-decrypt-badpass.log"
 decrypted_no_file_log="${s_basename}-decrypt-no-file.log"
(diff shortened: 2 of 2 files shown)


tokenjunkielabs:
Following up on first-reporter bug bounty #431. PR #432 is the `woahwhattheheck` fixing carrier for passphrase-file carriage-return truncation, where the program can silently use a passphrase different from the file contents. Please review the report/fix when convenient, classify #431 under the published Tarsnap bug-bounty schedule, and confirm the applicable payment route if accepted. I am not presuming a tier.

